PR #11, opened 2023-06-17
Base implementation of inspect request impl.

mvassilev:
There is a problem here with the inspect_request_impl function not being invoked which prevents handling input like "? std::vector" that is supposed to display documentation to the user.


vgvassilev:
@SylvainCorlay do you have a clue?

SylvainCorlay:
Hey there, sorry for not getting back to you earlier.

The way to trigger these inspect requests from a JupyterLab UI is to open the "Contextual Help" from the "Help" topbar menu.

Then any keystroke will trigger an inspect request when the contextual help is open.

ioanaif:
Hi! 

I have this working locally and will add some changes here (as discussed with @vgvassilev via Skype) as soon as I finish cleaning my tests up. (later today)

SylvainCorlay:
Hourray, this ia amazing!

Ping @loicgouarin for awareness.

vgvassilev:
Do we have a clang-format configuration for this project? If not, shall we use the standard one from llvm?

vgvassilev:
@SylvainCorlay what is the way to move forward here?

gouarin:
@mvassilev and @ioanaif thank you for your PR. It's a nice feature and it's great that it's working again! I'm really sorry for the delay and I will try to make a review soon. I also need to look at what @SylvainCorlay did in this project. Since I wrote xeus-cling, I haven't spent much time contributing to it. I will try to change that ;).

@vgvassilev we have no clang-format for this project for now but I can add one.

CMakeLists.txt
```diff
@@ -171,7 +171,7 @@ macro(xeus_cpp_set_common_options target_name)
     endif ()
 
     set_target_properties(${target_name} PROPERTIES
-        INSTALL_RPATH_USE_LINK_PATH TRUE
```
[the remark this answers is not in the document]
SylvainCorlay: (This seems to be somewhat unrelated to the rest of the changes.)

etc/xeus-cpp/tags.d/xtensor.json
```diff
@@ -0,0 +1,4 @@
+{
+    "url": "https://xtensor.readthedocs.io/en/latest/",
+    "tagfile": "xtensor.tag"
```
SylvainCorlay, line 3: I suggest we don't install the xtensor tag files as part of this PR, since it should probably be installed by xtensor itself.

src/xinterpreter.cpp
```diff
@@ -307,30 +324,38 @@ namespace xcpp
...
-
         std::string error_message;
         llvm::raw_string_ostream error_stream(error_message);
         // Attempt normal evaluation
+        
         try
-        {
+        {   std::string exp = R"(\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)";
+            std::regex re(R"((\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)(\.?)*$)");
+            auto inspect_request = is_inspect_request(code, re);
```
SylvainCorlay, line 334: An inspect request is a different type of message, triggered by keystrokes in the editor when the inspector panel is open.

I don't think we should call this function `is_inspect_request` but more like `is_help_magic`

src/xinterpreter.cpp
```diff
@@ -307,30 +324,38 @@ namespace xcpp
...
+            if (inspect_request.first)
+            {
+                inspect(inspect_request.second[0], kernel_res, *m_interpreter);
```
SylvainCorlay, line 337: Happy to discuss this. Maybe we should not rebind into the inspect function here.

src/xinterpreter.cpp
mvassilev: is_help_magic returns bool so what should be the logic here in the cases of true or false?

src/xinterpreter.cpp
mvassilev: Maybe this would be related to the previous conversation regarding the replacement of is_inspect_request with is_help_magic

CMakeLists.txt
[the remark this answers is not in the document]
vgvassilev: So in that case I'd propose to revert these changes.

vgvassilev:
@SylvainCorlay, @gouarin, this looks mostly okay to me. 

@SylvainCorlay, can you take a look and answer to the remaining questions?

I have a question on my own: Is there a way to avoid putting `share/xeus-cpp/tagfiles/cppreference-doxygen-web.tag` in the repository and download it from somewhere instead?

vgvassilev (approved):
I'd prefer to move forward here and break the stalemate. We can rely on a post-commit review here. LGTM!

SylvainCorlay:
> I have a question on my own: Is there a way to avoid putting share/xeus-cpp/tagfiles/cppreference-doxygen-web.tag in the repository and download it from somewhere instead?

I don't think this file exist otherwise.

PR #12, opened 2023-06-29
Update readthedocs configuration file

Krishna-13-cyber:
Build Status and Preview: https://readthedocs.org/projects/xeus-cpp-docs/builds/21147633/


docs/source/conf.py
```diff
@@ -9,17 +9,17 @@
...
 
-def setup(app):
-    app.add_stylesheet("main_stylesheet.css")
+# def setup(app):
+#     app.add_stylesheet("main_stylesheet.css")
 
-extensions = ['breathe']
-breathe_projects = { 'xeus-cpp': '../xml' }
+# extensions = ['breathe']
+# breathe_projects = { 'xeus-cpp': '../xml' }
```
SylvainCorlay, line 22: I think we should still keep the doxygen/breathe setup that we have in e.g. xeus-cling!

SylvainCorlay (commented):
Thank you for the contribution. I have left an inline comment.

docs/source/conf.py
Krishna-13-cyber: > I think we should still keep the doxygen/breathe setup that we have in e.g. xeus-cling!

Hi Sylvain,
I was unable to import the breathe setup when I was testing.
```Could not import extension breathe (exception: No module named 'breathe')```

I am using the command for build:
``` make html``` in xeus-cpp/docs/

When I link it in.readthedocs.yaml to environment.yml where we have the breathe dependency, it again shows importing error,
```
version: 2

sphinx:
  configuration: docs/source/conf.py
  builder: html

conda:
  environment: docs/environment.yml
  ```
Please let me know if I am on the right path!
Thanks.

Krishna-13-cyber: Hi Sylvain,
Pinging you for this reviewal!

Krishna-13-cyber: Hi @SylvainCorlay,
I am working on the docs setup for xeus-clang-repl but I am unsuccessful due to the issues below,
I was not able to ping you in the xeus-clang-repl (PR thread) due to some Github restrictions so I am pinging you in this thread itself!

So basically I had used cpp-zmq at the initial stage for xeus-clang-repl but they have updated to xeus-zmq for the implementation.
Now, I am using the conda-forge setup,
When I use xeus-zmq in my dependency list, with version 1.1.0 it doesn't turn out well in the readthedocs setup as it's not able to fetch the xeus=1.1.0, (https://readthedocs.org/api/v2/build/21405750.txt)
I am using xeus-zmq in my Cmakelists via find_package, and it gives me the xeus-zmq.config not found, (package not found).
I tried using pkgconfig for getting the xeus-zmq in Cmakelists via pkg_check_modules, not sure whether the right way of getting xeus-zmq like the the way we get zeromq.
Please let me know the right way to get the dependencies for xeus-clang-repl via conda-forge

Thanks!

vgvassilev: It looks like this is resolved now. Let's merge it to make sure that we have a configuration file v2 before the deadline of readthedocs.

PR #13, opened 2023-08-22
Update xeus-cpp documentation setup

Krishna-13-cyber:


CMakeLists.txt
```diff
@@ -106,9 +106,14 @@ if(XEUS_CPP_EMSCRIPTEN_WASM_BUILD)
...
-find_package(argparse REQUIRED)
+# find_package(argparse REQUIRED)
 find_package(pugixml REQUIRED)
-
+include(FetchContent)
+FetchContent_Declare(
+    argparse
+    GIT_REPOSITORY https://github.com/p-ranav/argparse.git
+)
+FetchContent_MakeAvailable(argparse)
```
vgvassilev, line 116: Why `find_package(argparse ...)` is not sufficient?

Krishna-13-cyber: Yes,
I tried fetching cppargparse as well argparse from conda but it did'nt find the dependency,
```
ResolvePackageNotFound: 
  - cpp-argparse
 ```
  leading to the following error
  ```
  Could not find a package configuration file provided by "argparse" with any
  of the following names:

    argparseConfig.cmake
    argparse-config.cmake
```

CMakeLists.txt
```diff
@@ -108,7 +108,12 @@ endif()
 find_package(Clang REQUIRED)
 find_package(argparse REQUIRED)
 find_package(pugixml REQUIRED)
-
+# include(FetchContent)
+# FetchContent_Declare(
+#     argparse
+#     GIT_REPOSITORY https://github.com/p-ranav/argparse.git
+# )
+# FetchContent_MakeAvailable(argparse)
```
vgvassilev, line 116: Please remove that commented out code.

vgvassilev (approved):
LGTM!

SylvainCorlay:
Thanks for this work.

I think it would be interesting to keep the directory structure that is common with all the xeus-based kernels for the documentation. For example, this broke the link to the logo in the readme.

vgvassilev:
@Krishna-13-cyber, could you look into this?

@SylvainCorlay, do we have a better place to that documentation?

SylvainCorlay:
I think it is all fine! My comment was merely that moving the source files from `docs/source` to `docs` makes this repo slightly different from others - and accedentaly broke the readme.

This also makes re-generating the cookie-cutter bits harder.

Krishna-13-cyber:
> I think it is all fine! My comment was merely that moving the source files from `docs/source` to `docs` makes this repo slightly different from others - and accedentaly broke the readme.
> 
> This also makes re-generating the cookie-cutter bits harder.
Hi,
So, should we make it back to the original convention of `docs/source`?

SylvainCorlay:
> So, should we make it back to the original convention of docs/source?

That would be great - just for keeping the xeus-based kernel as similar as possible where possible.

Krishna-13-cyber:
> > So, should we make it back to the original convention of docs/source?
> 
> That would be great - just for keeping the xeus-based kernel as similar as possible where possible.

Yes, I will get this done.
Thanks!

PR #17, opened 2023-11-08
Update xeus-cpp dependency from clangdev 16 to 17

anutosh491:
This PR fixes the **make && make install** build failures while using **clang 17.0.4** and **llvm 17.0.4**. The main changes here are inspired by (https://github.com/llvm/llvm-project/commit/b8e5f918166c46f8b3c0674f2a2e81bfcb95f969 and https://github.com/llvm/llvm-project/commit/8b1771bd9f304be39d4dcbdcccedb6d3bcd18200) as 
`ExecutorAddr` was introduced in  as an eventual replacement for `JITTargetAddress` and `ExecutorSymbolDef` was introduced as a 
replacement for `JITEvaluatedSymbol` in the following patches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ mamba install`xeus-cpp` notebook -c conda-forge
 Or you can install it from the sources, you will first need to install dependencies
 
 ```bash
-mamba install cmake cxx-compiler xeus-zmq nlohmann_json cppzmq xtl jupyterlab clangdev=16 cpp-argparse pugixml -c conda-forge
+mamba install cmake cxx-compiler xeus-zmq nlohmann_json cppzmq xtl jupyterlab clangdev=17 cpp-argparse pugixml -c conda-forge
 ```
 
 Then you can compile the sources (replace `$CONDA_PREFIX` with a custom installation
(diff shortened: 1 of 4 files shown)


vgvassilev:
Do you have any clue why things fail on osx?

anutosh491:
Looks like `clang_osx64 17` is not available yet ( latest is `16.0.6`). So we'll maybe have to wait until the switch happens to clang 17 on osx. @vgvassilev

anutosh491:
I think clang_osx 17 is now available (https://github.com/conda-forge/clang-compiler-activation-feedstock/pull/113) but we might have to wait for it to be enabled as the default compiler (https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/cbccb41e313c017c1ad15db6c287e87aa57d54dd/recipe/conda_build_config.yaml#L18)

vgvassilev:
LLVM 17 seems broken. It has valgrind issues discovered by CppInterOp. We should probably not update before fixing them: https://github.com/compiler-research/CppInterOp/actions/runs/7561682890/job/20590508932#step:26:11915

vgvassilev:
That's not needed anymore. It will be handled on the CppInterOp side. Do you mind closing this one?

anutosh491:
Yeah sure it makes sense. let's close this for now !

PR #18, opened 2023-11-14
Removed is_inspect_request call from execute request

anutosh491:
As discussed, we don't really need an `is_inspect_request` call inside the execute request implementation. Which ends up giving us something like 
image

Now that has been addressed 
image

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -327,16 +327,8 @@ namespace xcpp
         // Attempt normal evaluation
         
         try
-        {   std::string exp = R"(\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)";
-            std::regex re(R"((\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)(\.?)*$)");
-            auto inspect_request = is_inspect_request(code, re);
-            if (inspect_request.first)
-            {
-                inspect(inspect_request.second[0], kernel_res, *m_interpreter);
-            }
-            
+        {
             compilation_result = process_code(*m_interpreter, code, error_stream);
-        
         }
         catch (std::exception& e)
         {
@@ -427,7 +419,7 @@ namespace xcpp
     {
         nl::json kernel_res;
         std::string exp = R"(\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)";
-        std::regex re(R"((\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)(\.?)*$)");
+        std::regex re{"(" + exp + R"(\.?)*$)"};
         auto inspect_request = is_inspect_request(code.substr(0, cursor_pos), re);
         if (inspect_request.first)
         {


src/xinterpreter.cpp
```diff
@@ -327,16 +327,8 @@ namespace xcpp
...
         try
-        {   std::string exp = R"(\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)";
-            std::regex re(R"((\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)(\.?)*$)");
-            auto inspect_request = is_inspect_request(code, re);
-            if (inspect_request.first)
-            {
-                inspect(inspect_request.second[0], kernel_res, *m_interpreter);
-            }
-            
+        {
```
ioanaif, line 330: The `is_inspect_request` needs to parse the code, before `process_code` and catch the cases where we have a request such as: `?std::vector`

The [`no documentation printout`](https://github.com/compiler-research/xeus-cpp/blob/7c7d42de1cc40b595418d8a675c6648ea7dc1860/src/xinspect.hpp#L220) could be fixed with a better regex that only matches when we have a `?` to check for documentation

src/xinterpreter.cpp
```diff
@@ -427,7 +419,7 @@ namespace xcpp
     {
         nl::json kernel_res;
         std::string exp = R"(\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)";
-        std::regex re(R"((\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)(\.?)*$)");
+        std::regex re{"(" + exp + R"(\.?)*$)"};
         auto inspect_request = is_inspect_request(code.substr(0, cursor_pos), re);
```
ioanaif, line 423: In the case of `?std::vector` (as an example), the kernel will have shut down with an error in `process_code` before reaching this line

src/xinterpreter.cpp
anutosh491: Ohh so an execution request and an inspect request aren't decoupled. As in if I want to inspect `?std::vector` , shouldn't that be calling the `inspect_request_impl` ? 

I thought that something like the above would not end up calling `execute_request_impl`. I don't see close relatives of xeus-cpp (like xeus-cling and xeus-clang-repl) indulge in any kind of inspection in the execute request code (https://github.com/jupyter-xeus/xeus-cling/blob/471b1b68e3cad35746652cf14ce40ce0b7797ba0/src/xinterpreter.cpp#L125, https://github.com/compiler-research/xeus-clang-repl/blob/54d7f2c1d373855525ab2e9b86da3164376d7e1f/src/xinterpreter.cpp#L113) 

Could you explain this to me ? 
cc @ioanaif

src/xinterpreter.cpp
anutosh491: I made this change considering the variable `exp` was defined but not being used throughout the codebase and I think the purpose of defining it was so that it could be used for the regex.
But after going through your comment above, I am not sure what prompt should even trigger this function. 
I thought something like `?std::vector` would end up triggering this rather than entering the execute request code block.

ioanaif:
Since you have this set up, could you please test if after your changes `?std::vector` pops up the documentation without crashing the kernel? If it does then indeed, `inspect_request_impl` (1) and `execute_request_impl`(2) are decoupled correctly (as they should be). And all is good. 

I was under the impression that the check for `is_inspect_request` before L338 was put there because the decoupling of the two (1,2) was not done properly; However, if the test above passes, then everything is fine and your changes take out useless code duplication.

PR #19, opened 2023-11-16
Fixup wasm

SylvainCorlay:
The llvm package for emscripten-forge includes clang.


anutosh491:
Was addressed through #22 , hence closing !

PR #21, opened 2023-11-27
Fixed inspect request using Preamble approach

anutosh491:
On **main** 

image

On **branch**

image

### DIFF
--- a/src/xinspect.hpp
+++ b/src/xinspect.hpp
@@ -160,6 +160,7 @@ namespace xcpp
                     tagfile = it->at("tagfile");
                     std::string filename = tagfiles_dir + "/" + tagfile;
                     pugi::xml_document doc;
+                    doc.load_file(filename.c_str());
                     class_member_predicate predicate{typename_, "function", method[2]};
                     auto node = doc.find_node(predicate);
                     if (!node.empty())
@@ -193,6 +194,7 @@ namespace xcpp
                 tagfile = it->at("tagfile");
                 std::string filename = tagfiles_dir + "/" + tagfile;
                 pugi::xml_document doc;
+                doc.load_file(filename.c_str());
                 for (auto c : check)
                 {
                     node_predicate predicate{c, find_string};
@@ -262,5 +264,36 @@ namespace xcpp
             kernel_res["status"] = "ok";
         }
     }
+
+    class xintrospection : public xpreamble
+    {
+    public:
+
+        using xpreamble::pattern;
+        const std::string spattern = R"(^\?)";
+
+        xintrospection(clang::Interpreter& p)
+            : m_interpreter{p}
+        {
+            pattern = spattern;
+        }
+
+        void apply(const std::string& code, nl::json& kernel_res) override
+        {
+            std::regex re(spattern + R"((.*))");
+            std::smatch to_inspect;
+            std::regex_search(code, to_inspect, re);
+            inspect(to_inspect[1], kernel_res, m_interpreter);
+        }
+
+        virtual xpreamble* clone() const override
+        {
+            return new xintrospection(*this);
+        }
+
+    private:
+
+        clang::Interpreter& m_interpreter;
+    };
 }
 #endif
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,0 +1,69 @@
+#############################################################################
+# Copyright (c) 2023, xeus-cpp contributors                                 #
+#                                                                           #
+# Distributed under the terms of the BSD 3-Clause License.                  #
+#                                                                           #
+# The full license is in the file LICENSE, distributed with this software.  #
+#############################################################################
+
+
+# Unit tests
+# ==========
+
+cmake_minimum_required(VERSION 3.1)
+
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    project(xeus-cpp-test)
+
+    enable_testing()
+
+    find_package(xeus-cpp REQUIRED CONFIG)
+endif()
+
+include(CheckCXXCompilerFlag)
+
+string(TOUPPER "${CMAKE_BUILD_TYPE}" U_CMAKE_BUILD_TYPE)
+
+if(CMAKE_CXX_COMPILER_ID MATCHES Clang OR CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Intel)
+    add_compile_options(-Wunused-parameter -Wextra -Wreorder -Wconversion -Wsign-conversion)
+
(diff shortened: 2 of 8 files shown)


anutosh491:
cc @ioanaif 
Continuing our discussion from Friday. I have the following 

After talking to Johan, I got to know about the following 

i) The execute request and the inspect request are decoupled but then at the same time as use `shift + enter` to trigger both of them, the inspect request for a client like JupyterLab would be processed through an execution request only
ii) But then what role does `inspect_request_impl` play ? It comes into picture when other client like a jupyter console is used .Hence we have established that we need an inspection check separately and also one through the execution request.
iii) But should we use the inspect function in the `execute_request_impl` body ? After discussing with Johan I realized that a better method to do this would be using preambles. This helps us overall (also when we get to magics in the near future) .

So we have the following 
A cell with a behavior defined in the preamble manager which depends on a certain pattern. For example:
1) if a cell begins with %% or %, it is a magic cell or magic line,
2) if a cell begins with !, it is a shell command,
3) if a cell begins with ?, it is an introspection.

vgvassilev:
Is there a way to add tests to this PR?

cc: @JohanMabille

vgvassilev:
@anutosh491, could you do that?

test/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+        add_compile_options(-march=native)
+    endif()
+endif()
+
+if(CMAKE_CXX_COMPILER_ID MATCHES MSVC)
+    add_compile_options(/EHsc /MP /bigobj)
+    set(CMAKE_EXE_LINKER_FLAGS /MANIFEST:NO)
+endif()
+
+find_package(doctest)
```
vgvassilev, line 44: Let’s use gtest which seems standard for the projects that are close to llvm.

JohanMabille: I would prefer to keep doctest. gtest had a period with almost no maintainance (I had to keep a forked branch for more than one year before they merged and released a patch I submitted, and the bug prevented from testing xtensor ), and had a bunch of ABI compatibilities issues (we had to build it from source when trying to test with different compilers). Besides, doctest is used for the whole xeus stack now.

JohanMabille (commented):
I would add a `test_interpreter.cpp` file instead of `test_execute_request`, for:
- granularity, we don't want one file per tested feature
- naming: `test_execute_request` sounds like you test the whole path of an execute request sent from a client (including the middleware part)

src/xinspect.hpp
```diff
@@ -160,6 +160,7 @@ namespace xcpp
                     tagfile = it->at("tagfile");
                     std::string filename = tagfiles_dir + "/" + tagfile;
                     pugi::xml_document doc;
+                    pugi::xml_parse_result result __attribute__((unused)) = doc.load_file(filename.c_str());
```
JohanMabille, line 163: `__attribute__` is UNIX only, it should be avoided (or used through macros with different definitions depending on the platform). If you don't need `result`, you can simply call `doc.load_file` without assigning its result, or use the `(void*)result;` trick to avoid the warning.

test/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+        BUILD_WITH_INSTALL_RPATH 1
+        SKIP_BUILD_RPATH FALSE
+    )
+endif()
+
+set_target_properties(test_xeus_cpp PROPERTIES
+    INSTALL_RPATH_USE_LINK_PATH TRUE
+)
+
+target_link_libraries(test_xeus_cpp xeus-zmq doctest::doctest ${CMAKE_THREAD_LIBS_INIT})
```
JohanMabille, line 69: You need to link with `xeus-cpp` (or `xeus-cpp-static`), not `xeus-zmq`

test/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    project(xeus-cpp-test)
+
+    enable_testing()
+
+    find_package(xeus-cpp REQUIRED CONFIG)
+endif()
+
+message(STATUS "Forcing tests build type to Release")
+set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
```
vgvassilev, line 24: Why do we need to force the tests to be in release while everything else has been selected to be in Debug, for example? That might cause build inconsistencies when we have `#ifndef NDEBUG`, etc.

test/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+
+    CHECK_CXX_COMPILER_FLAG(-march=native HAS_MARCH_NATIVE)
+    if (HAS_MARCH_NATIVE)
+        add_compile_options(-march=native)
+    endif()
+endif()
+
+if(CMAKE_CXX_COMPILER_ID MATCHES MSVC)
+    add_compile_options(/EHsc /MP /bigobj)
+    set(CMAKE_EXE_LINKER_FLAGS /MANIFEST:NO)
```
vgvassilev, line 41:
```suggestion
    add_link_options(/MANIFEST:NO)
```

test/main.cpp
```diff
@@ -0,0 +1,10 @@
+/***************************************************************************
+ * Copyright (c) 2023, xeus-cpp contributors
+ *
+ * Distributed under the terms of the BSD 3-Clause License.
+ *
+ * The full license is in the file LICENSE, distributed with this software.
+ ****************************************************************************/
+
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include "doctest/doctest.h"
```
vgvassilev, line 10: Why do we need this file? Can we move its contents in `test_interpreter`?

anutosh491: Ahh maintaing consistency across all xeus kernels. xeus-kernels like xeus-python & xeus-cling use doctests through a main.cpp file.

src/xinspect.hpp
```diff
@@ -160,6 +160,7 @@ namespace xcpp
                     tagfile = it->at("tagfile");
                     std::string filename = tagfiles_dir + "/" + tagfile;
                     pugi::xml_document doc;
+                    static_cast<void>(doc.load_file(filename.c_str()));
```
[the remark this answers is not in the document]
anutosh491: We don't ! Thanks for pointing this out.

test/CMakeLists.txt
anutosh491: That does make sense. We can get rid of it.

vgvassilev (approved):
LGTM! We should add a subsequent PR tracking the testing coverage. We can take inspiration from https://github.com/compiler-research/CppInterOp/blob/d294b0790318522fd1eb0b0a2c786a4ba0347a58/.github/workflows/ci.yml#L46

PR #22, opened 2023-11-28
Attempt to fix wasm build

anutosh491:


include/xeus-cpp/xinterpreter_wasm.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef XEUS_CPP_INTERPRETER_WASM_HPP
+#define XEUS_CPP_INTERPRETER_WASM_HPP
+
+#include "xinterpreter.hpp"
+#include "xeus_cpp_config.hpp"
+
+namespace xcpp
+{
+    class XEUS_CPP_API wasm_interpreter : public interpreter
```
vgvassilev, line 19: Is there a particular reason we call this class a `wasm_interpreter`? Does it do anything wasm-specific?

anutosh491: Yes, we use this for the wasm build. 

Basically if you check out the file `src/main_emscripten_kernel.cpp`. There we are trying to use the `export_kernel` function out of xeus (https://github.com/jupyter-xeus/xeus-lite/blob/e6438a2e830300716e13abdb98724ec68435767e/include/xeus-lite/xembind.hpp#L63) which calls another function `make_xkernel` which expects the `interpreter_type`

But what we can't do is pass something like `xcpp::interpreter(interpreter_argc, interpreter_argv)` for the interpreter_type because kernels / interpreters do not have an API to pass parameters like that. So there needs to be a default for whatever interpreter_argc, interpreter_argv is suppose to be
In the jupyter(lite) ui there would not be a way to pass these arguments...so this just wont work.

For that what we do is use a `wasm_interpreter` derived from the base `interpreter`
```
    wasm_interpreter::wasm_interpreter()
        : interpreter(0, nullptr)
```
And in the constructor assign default values to `argc` and `argv` (in file `src/xinterpreter_wasm.cpp`) and then use this `wasm_interpreter` instead of the usual one.

This approach is quite consistent among other xues kernels like xeus-python and xeus-lua. You can have a look here (https://github.com/jupyter-xeus/xeus-python/pull/530)

vgvassilev: Ok, I think I understand now. Thanks for clarifying.

PR #25, opened 2024-01-12
Fix extract_filename function

anutosh491:
If the `extract_filename` function is going to be editing `argv`, then it also needs to make a corresponding change to `argc`. Can lead to inconsistencies otherwise.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -27,81 +27,11 @@
 
 #include "xeus-cpp/xeus_cpp_config.hpp"
 #include "xeus-cpp/xinterpreter.hpp"
-
-#ifdef __GNUC__
-void handler(int sig)
-{
-    void* array[10];
-
-    // get void*'s for all entries on the stack
-    std::size_t size = backtrace(array, 10);
-
-    // print out all the frames to stderr
-    fprintf(stderr, "Error: signal %d:\n", sig);
-    backtrace_symbols_fd(array, size, STDERR_FILENO);
-    exit(1);
-}
-#endif
-
-void stop_handler(int /*sig*/)
-{
-    exit(0);
-}
-
-bool should_print_version(int argc, char* argv[])
-{
-    for (int i = 0; i < argc; ++i)
-    {
-        if (std::string(argv[i]) == "--version")
-        {
-            return true;
-        }
-    }
-    return false;
-}
-
-std::string extract_filename(int argc, char* argv[])
-{
-    std::string res = "";
-    for (int i = 0; i < argc; ++i)
-    {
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -60,19 +60,19 @@ bool should_print_version(int argc, char* argv[])
     return false;
 }
 
-std::string extract_filename(int argc, char* argv[])
+std::string extract_filename(int *argc, char* argv[])
```
JohanMabille, line 63: Prefer passing references instead of pointers when you can (which is the case for argc, not for argv)

JohanMabille (changes requested):

src/main.cpp
anutosh491: Argh argh argh yes fundamental mistake. Overlooked it. Thanks !

vgvassilev: Can we add a test for this?

anutosh491: Hmm not sure, how we could test this out :(
Just something I noticed which I thought I'll address!

JohanMabille: You can forward declare the function in the test file, the definition will be found when linking.

JohanMabille (approved):

src/main.cpp
anutosh491: Yeah that makes sense. Done.

test/test_interpreter.cpp
```diff
@@ -9,6 +9,9 @@
 #include "doctest/doctest.h"
 #include "xeus-cpp/xinterpreter.hpp"
 
+// Forward declaration of extract_filename
+std::string extract_filename(int &argc, char* argv[]);
```
vgvassilev, line 13: Now we need to call this function with different arguments and make sure it does what’s expected.

anutosh491: Okay I'll add a basic test from what comes to my mind.

src/xutils.cpp
```diff
@@ -0,0 +1,85 @@
...
+                break;
+            }
+        }
+        return res;
+    }
+
+    interpreter_ptr build_interpreter(int argc, char** argv)
+    {
+        int interpreter_argc = argc;  // + 1;                                               // ...
+        const char** interpreter_argv = new const char*[interpreter_argc];
```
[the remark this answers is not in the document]
anutosh491: Cool I think we can look into any more refactoring through subsequent PRs. The PR has already deivated a bit from what it intended to do. I'll raise one to address the `build_interpreter` function and relevant tests next.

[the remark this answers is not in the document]
vgvassilev: Fair enough.

PR #30, opened 2024-01-18
[clang-tidy] Adjust xeus-cpp build invocation.

vgvassilev:


vgvassilev:
@alexander-penev, I will need help how to install the dependencies of xeus-cpp using apt-get so that we can configure xeus-cpp inside the clang-tidy docker container. Do you have an idea how to do that?

PR #36, opened 2024-03-11
Add test cases for xutils, xmagics_manager, xpreamble_manager and xbuffer

brauliorivas:
This pull request adds test cases for a few functions in the xcpp library.

`should_print_version` Test Suite: Tests verify if the function correctly identifies the presence or absence of the `--version` argument.

`build_interpreter` Test Suite: A test case checks if the function returns a non-null pointer when valid arguments are passed.

`clone_magics_manager` Test Suite: Tests ensure the function correctly returns a non-null pointer and that the cloned object matches the type of the original magics_manager.

`is_match` Test Suite: Verifies the function correctly identifies strings that match the regex pattern used in `xmagics_manager`.

`register_and_access` Test Suite: This test verifies the correct registration and access of a `xpreamble` object within the `xpreamble_manager`. It ensures that a `xpreamble` object can be registered with a name and subsequently accessed using the same name.

`xbuffer` Test Suite: The new test cases cover the `xinput_buffer`, `xoutput_buffer`, and `xnull` classes, focusing on their interaction with various inputs and outputs. They verify the correct behavior of these classes when handling normal and edge cases, such as empty outputs, the correct invocation of callback functions, and the discarding of output by `xnull`.

Currently, these new test cases were executed and successfully passed.

### DIFF
--- a/test/test_interpreter.cpp
+++ b/test/test_interpreter.cpp
@@ -21,7 +21,13 @@ TEST_SUITE("execute_request")
         std::string inspect_result = "https://en.cppreference.com/w/cpp/container/vector";
         nl::json user_expressions = nl::json::object();
 
-        nl::json result = interpreter.execute_request(code, false, false, user_expressions, false);
+        nl::json result = interpreter.execute_request(
+            code, 
+            false, 
+            false, 
+            user_expressions, 
+            false
+        );
 
         REQUIRE(result["payload"][0]["data"]["text/plain"] == inspect_result);
         REQUIRE(result["user_expressions"] == nl::json::object());
--- a/test/test_interpreter.cpp
+++ b/test/test_interpreter.cpp
@@ -8,6 +8,7 @@
 
 #include "doctest/doctest.h"
 #include "xeus-cpp/xinterpreter.hpp"
+#include "xeus-cpp/xholder.hpp"
 #include "xeus-cpp/xmanager.hpp"
 #include "xeus-cpp/xutils.hpp"
 
--- a/test/test_interpreter.cpp
+++ b/test/test_interpreter.cpp
@@ -8,26 +8,20 @@
 
 #include "doctest/doctest.h"
 #include "xeus-cpp/xinterpreter.hpp"
+#include "xeus-cpp/xmanager.hpp"
 #include "xeus-cpp/xutils.hpp"
 
 TEST_SUITE("execute_request")
 {
     TEST_CASE("fetch_documentation")
     {
-
         xcpp::interpreter interpreter(0, nullptr);
 
         std::string code = "?std::vector";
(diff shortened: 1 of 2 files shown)


test/test_interpreter.cpp
```diff
@@ -48,3 +48,232 @@ TEST_SUITE("extract_filename")
...
+        char* argv[] = {arg1, arg2};
+        int argc = 2;
+
+        interpreter_ptr interp_ptr = xcpp::build_interpreter(argc, argv);
+
+        REQUIRE(interp_ptr != nullptr);
+    }
+}
+
+TEST_SUITE("is_match_magics_manager")
```
maximusron, line 104: @brauliorivas if I understand correctly you can improve this test to check `xmagics_manager::apply` with line magic instead. it would std::regex_search %python same as is_match but also check if the magic can be successfully applied(and possibly record any breaking of the magic commands)

brauliorivas: alright! I'll change it

maximusron:
@anutosh491 could you take a look at this PR?
thanks!

test/test_interpreter.cpp
brauliorivas: Hey @maximusron I'm working on this test case as you said but I'm having trouble. If you have time, some help would be handy!

vgvassilev:
@brauliorivas, do you have any clue which part of the new changes breaks windows?

brauliorivas:
> @brauliorivas, do you have any clue which part of the new changes breaks windows?

I'm going to debug and try to solve the Windows issue.

test/test_interpreter.cpp
```diff
@@ -48,3 +49,232 @@ TEST_SUITE("extract_filename")
...
+    // a `xpreamble` object. It sets up a scenario where a `xpreamble` object is registered
+    // with a name and checks if the object can be accessed using the same name.
+    TEST_CASE("register_and_access")
+    {
+        std::string name = "test";
+        xcpp::xpreamble_manager manager;
+        xcpp::xmagics_manager* magics = new xcpp::xmagics_manager();
+        manager.register_preamble(name, magics);
+
+        xcpp::xholder_preamble& result = manager.operator[](name);
```
vgvassilev, line 179: We should probably annotate with `XEUS_CPP_API` the `xholder_preamble` class...

brauliorivas: Alright, I've added this

brauliorivas: Hey @vgvassilev, thank you so much for the help, it worked!

vgvassilev (commented):
LGTM! However, I'd wait for @anutosh491's review.

vgvassilev:
> LGTM! However, I'd wait for @anutosh491's review.

I'd like to move forward and rely on a post-commit review here.

PR #50, opened 2024-03-19
Add a test making sure we have set up STL properly

vgvassilev:


vgvassilev:
Ok, that's broken today, we should keep it as part of #46...

PR #53, opened 2024-03-21
Pin nlohmann_json to the last working version

anutosh491:
Trying to fix latest build failures on main


vgvassilev (approved):
Awesome! LGTM!

PR #58, opened 2024-03-25
Merge xeus-clang-repl mini - new kernels

alexander-penev:


mcbarton:
@alexander-penev What is the difference between the cpp 17 (xcpp) kernel and the C++17 (xcpp) kernel? They appear to be the same.

alexander-penev:
> @alexander-penev What is the difference between the cpp 17 (xcpp) kernel and the C++17 (xcpp) kernel? They appear to be the same.

Yes, maybe we should remove "cpp 17 (xcpp)".

CMakeLists.txt
```diff
@@ -126,13 +126,38 @@ if (NOT DEFINED XEUS_CPP_KERNELSPEC_PATH)
...
 endif ()
 
-set(XEUS_CPP_PATH "$ENV{PATH}")
-set(XEUS_CPP_LD_LIBRARY_PATH "$ENV{LD_LIBRARY_PATH}")
-set(XEUS_CPP_RESOURCE_DIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/clang/${CppInterOp_CLANG_VERSION})
-configure_file (
-    "${CMAKE_CURRENT_SOURCE_DIR}/share/jupyter/kernels/xcpp/kernel.json.in"
-    "${CMAKE_CURRENT_BINARY_DIR}/share/jupyter/kernels/xcpp/kernel.json"
-)
+function(configure_kernel kernel)
```
JohanMabille, line 129: This function is mixing configuration and install. This should be splitted:
- The different `kernel.json.in` files must be configured to produce `kernel.json` files in the source directory
- Another function should be implemented to copy the different files to the binary dir (in the install section of the CMakeLists.txt)

CMakeLists.txt
```diff
@@ -126,13 +126,38 @@ if (NOT DEFINED XEUS_CPP_KERNELSPEC_PATH)
...
+  set(XEUS_CPP_PATH "$ENV{PATH}")
+  set(XEUS_CPP_LD_LIBRARY_PATH "$ENV{LD_LIBRARY_PATH}")
+  set(XEUS_CPP_RESOURCE_DIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/clang/${CppInterOp_CLANG_VERSION})
+
+  configure_file (
+    "${CMAKE_CURRENT_SOURCE_DIR}/${kernel}/kernel.json.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/${kernel}/kernel.json"
+    )
+  file(GLOB files "${CMAKE_CURRENT_SOURCE_DIR}/${kernel}/*.png")
```
JohanMabille, line 138: Explicitly listing the files is considered as a better practice than `file(GLOB)`.

CMakeLists.txt
```diff
@@ -126,13 +126,38 @@ if (NOT DEFINED XEUS_CPP_KERNELSPEC_PATH)
...
+      )
+  endforeach()
+  file(GLOB files "${CMAKE_CURRENT_SOURCE_DIR}/${kernel}/*.svg")
+  foreach(file ${files})
+    configure_file(
+      "${file}"
+      "${CMAKE_CURRENT_BINARY_DIR}/${kernel}/"
+      COPYONLY
+      )
+  endforeach()
```
JohanMabille, line 153: No need to loop over the files and call configure. Once you've caputred all the files in the `files`variable, you can simply call `install` to copy the files (see [here](https://github.com/jupyter-xeus/xeus-python/blob/main/CMakeLists.txt#L453-L455) for an example).

JohanMabille (changes requested):
Do we need to keep `share/jupyter/kernels/kernel.json.in` ?

CMakeLists.txt
vgvassilev: Why we should produce the `kernel.json` in the source directory? The build process should not touch the source directories. That's the recommended way of cmake but in addition the source directory can be on read-only file system. What do we gain if we output the `kernel.json` in the source directory?

JohanMabille: This was for consistency with the cmake files of other xeus projects; but you're right, we should not modify the source directory.

JohanMabille (approved):

CMakeLists.txt
vgvassilev: Ok, understood. Thanks for reviewing!

Issue #42, opened 2024-03-14
[ci] Remove the duplication in the ci matrix


vgvassilev:
We can invert the CI matrix to include only the different options and the steps can become the same. For reference see how's done in the vgvassilev/clad project.

vgvassilev:
@mcbarton, is that something that you can do for 3 minutes with your ever-increasing CI skills ;)

PR #63, opened 2024-05-03
Reduce repetition ci

mcbarton:
This PR removes duplication within the ci
Fixes https://github.com/compiler-research/xeus-cpp/issues/42

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  unix:
+  build:
     
     runs-on: ${{ matrix.os }}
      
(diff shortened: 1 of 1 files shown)


mcbarton:
@vgvassilev This PR is ready for review. I don't know where it specifically addresses what you wanted in the issue, but I have tried to remove repetition in the ci.

vgvassilev:
Nice! Any chance to incorporate the wasm build? It seems quite different...

mcbarton:
> Nice! Any chance to incorporate the wasm build? It seems quite different...

It may be possible, but I'd advise against it. I think the ci is clearer with the wasm and non wasm builds separate.

vgvassilev:
> > Nice! Any chance to incorporate the wasm build? It seems quite different...
> 
> It may be possible, but I'd advise against it. I think the ci is clearer with the wasm and non wasm builds separate.

Ok, makes sense. Let's get this one in.

PR #65, opened 2024-05-04
Update actions ci to remove warnings

mcbarton:
This PR should remove all warnings outputted by the ci running, due to outdated actions.


mcbarton:
@vgvassilev All actions have been updated (checkout, codecov-action and provision-with-micromamba) so that all warnings are gone when the ci runs. I consider this PR ready for review.

PR #67, opened 2024-05-07
Fix clang tidy

mcbarton:
@vgvassilev This is the PR to correct the bug introduced in clang tidy workflow in the last PR,  found in https://github.com/vgvassilev/clad/pull/883 . Once it is has been found to work in Clad this can be merged immediately.

### DIFF
--- a/.github/workflows/clang-tidy-review.yml
+++ b/.github/workflows/clang-tidy-review.yml
@@ -14,14 +14,20 @@ concurrency:
 jobs:
   review:
     runs-on: ubuntu-latest
+    
     steps:
       - name: Checkout PR branch
         uses: actions/checkout@v4
 
-      - name: Install LLVM and Clang
-        uses: KyleMayes/install-llvm-action@v2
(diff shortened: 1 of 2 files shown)


mcbarton:
@maximusron any ideas of how to fix the clang tidy workflow here? The remaining issue which I need to solve is how activate the xeus-cpp envirnoment before the cmake in ZedThree/clang-tidy-review , but any command you will put in the cmake_command will be prepended with a cmake command. I'm running out of ideas of how to have the xeus-cpp envirnoment activated.

mcbarton:
@vgvassilev @anutosh491 can one of you merge this PR? It fixes the issue of the broken clang tidy introduced by the previous commit. It works around an issue where an action tries to stop you using anything but a cmake command as the first command in the updated action. That's the only way I could find to activate the mamba envirnoment in the updated action. This PR also shows the clang tidy now works.

anutosh491 (approved):
Looks good. Merging !

PR #73, opened 2024-05-10
Add conda package manager information to readme

mcbarton:
This removes the line saying the package is not available within the conda package manager and replaces it with the badges showing its availibility. @alexander-penev @vgvassilev This PR is ready.


vgvassilev (commented):
LGTM!

PR #75, opened 2024-05-10
non hardcoded number of parallel jobs

mcbarton:
Previously the number of parallel jobs was hardcoded independent of the runner. This sets it dependent on the number of available cores.


mcbarton:
@vgvassilev can you merge this PR?

PR #77, opened 2024-05-10
Add c++20 kernel

mcbarton:
The purpose of this PR is to add a c++20 kernel.


alexander-penev (approved):

CMakeLists.txt
```diff
@@ -154,6 +154,8 @@ configure_kernel("share/jupyter/kernels/xcpp")
 configure_kernel("share/jupyter/kernels/xcpp11")
 configure_kernel("share/jupyter/kernels/xcpp14")
 configure_kernel("share/jupyter/kernels/xcpp17")
+configure_kernel("share/jupyter/kernels/xcpp17-omp")
```
vgvassilev, line 157: I am not sure I understand how this works. We need to pass somehow `-fopenmp` as a flag. `XEUS_CPP_OMP` probably does not expand to anything...

mcbarton: I'll remove adding the omp-17 kernel and move that to a different PR. I assumed that infrastructure was there to use it, but had accidentally been missed. I'll leave this PR to just adding the c++17 kernel.

mcbarton: > I'll remove adding the omp-17 kernel and move that to a different PR. I assumed that infrastructure was there to use it, but had accidentally been missed. I'll leave this PR to just adding the c++20 kernel.

@vgvassilev I've made this change, so the PR should hopefully be ok to merge now.

vgvassilev (commented):
LGTM. I'd wait for @JohanMabille.

PR #81, opened 2024-05-11
Fix available conda platforms

mcbarton:
The platforms for which xeus-cpp is available on Conda is not updating to showing the addition of Windows. This PR fixes that.


vgvassilev (approved):
Thank you, @mcbarton!

PR #83, opened 2024-05-11
Update example notebook to use c++17 kernel

mcbarton:
Previously the example notebook made use of the C++14 kernel . This PR updates it to use the C++17 kernel.


vgvassilev (approved):
LGTM!

PR #84, opened 2024-05-12
Revert required xeus version

alexander-penev:


JohanMabille:
Why this change?

vgvassilev:
I think it was in the context of trying out packaging xeus-cpp in https://github.com/emscripten-forge/recipes/pull/937 We should have done that in a branch but in the end it worked out. Does that answer your question or I am misunderstanding it?

JohanMabille:
Yes but we cannot keep it as is, otherwise xeus-cpp won't be installable with latest version of other xeus kernels; the xeus dependency should be upgraded to xeus 3.2. If the wasm build fails, then we should investigate why and fix it.

PR #87, opened 2024-05-12
Remove kernelspec from .gitignore

mcbarton:
Previously xeus-cpp only had one kernel. This PR updates the .gitignore to take into account there are now multiple kernels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,9 +38,6 @@
 # Python
 __pycache__/
 
-# Generated kernelspec
-share/jupyter/kernels/xcpp/kernel.json
-
 # Build directory
 build/
 


.gitignore
```diff
@@ -39,7 +39,7 @@
 __pycache__/
 
 # Generated kernelspec
-share/jupyter/kernels/xcpp/kernel.json
+share/jupyter/kernels/**/*.json
```
vgvassilev, line 42: Why is this even in the ignore list?

mcbarton: @vgvassilev I actually don't know. I assummed it was there for a reason, and generalised it. When I am next at a computer I will remove it, and change the PR name accordingly.

vgvassilev: I think we used to generate some files in the source directory with cmake configure_file. That’s gone now.

mcbarton:
@vgvassilev This PR is ready for merging.

PR #88, opened 2024-05-12
Treat all warnings as errors (non wasm)

mcbarton:
The PR will change it so when building xeus-cpp it treats all warnings as errors by default. In the ci I turn this off for wasm builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.  #
 #############################################################################
 
-cmake_minimum_required(VERSION 3.4.3)
+cmake_minimum_required(VERSION 3.24)
 project(xeus-cpp)
 
 enable_language(CXX)
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -101,8 +101,8 @@ int __get_cxx_version () {
 __get_cxx_version ()
       )";
 
-      long cxx_version = Cpp::Evaluate(code);
-      return std::to_string(cxx_version);
+        auto cxx_version = Cpp::Evaluate(code);
+        return std::to_string(cxx_version);
     }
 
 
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -47,6 +47,8 @@ OPTION(XEUS_CPP_BUILD_EXECUTABLE "Build the xcpp executable" ON)
 OPTION(XEUS_CPP_USE_SHARED_XEUS "Link xcpp  with the xeus shared library (instead of the static library)" ON)
 OPTION(XEUS_CPP_USE_SHARED_XEUS_CPP "Link xcpp  with the xeus shared library (instead of the static library)" ON)
 
+option(CMAKE_COMPILE_WARNING_AS_ERROR "Compile warnings as errors" ON)
```
JohanMabille, line 50: - No need to create an option here, `CMAKE_COMPILE_WARNING_AS_ERROR` already exists as a [cmake variable](https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html) and can be passed directly in the command line.
- Although it's a good thing to have it enabled by default on the CI, it should be off by default for other users
- The required minimal version of cmake should be updated to be compatible with the use of this variable.

src/xinterpreter.cpp
```diff
@@ -138,6 +138,9 @@ __get_cxx_version ()
     {
         nl::json kernel_res;
 
+
+        auto input_guard = input_redirection(allow_stdin);
```
JohanMabille, line 142: Why adding this line in this PR? This seems unrelated.

JohanMabille (changes requested):

src/xinterpreter.cpp
mcbarton: This is to fix a warning which comes when building xeus-cpp. Without it allow_stdin is an unused variable. I coped how its used in xeus-python.

CMakeLists.txt
mcbarton: This comment has now been addressed.

mcbarton:
@JohanMabille This PR is ready for another review. I have addressed your comments.

JohanMabille:
Thanks!

JohanMabille (approved):

PR #92, opened 2024-05-13
Fix it so jupyter_kernel_test runs without error on Windows

mcbarton:
This PR will make it so the jupyter kernel tests can run on Windows.


mcbarton:
The issue appears to be that when running the python tests on Windows that the kernel json cannot be parsed in order to do the tests if I read the output correctly. I will investigate.

mcbarton:
Tests are now running. Just need to work out why they are failing.

mcbarton:
@JohanMabille @vgvassilev can you review this PR? It makes changes such that the kernels can be parsed on Windows for the jupyter_kernel_test , so they can now run. I have stopped the failing tests from running on Windows, and raised it as an issue here https://github.com/compiler-research/xeus-cpp/issues/95.

PR #99, opened 2024-05-15
Update testing instructions

mcbarton:
This updates various files so they all contain the instructions to test the kernel. I also change it so its obvious that the build instructions are currently for the non wasm build. A PR to follow will add instructions on how to wasm build instructions.


mcbarton:
@vgvassilev This PR is ready for review.

alexander-penev (approved):

PR #103, opened 2024-05-16
Upgraded to latest xeus (3.2.0)

### LINKED ISSUE #94: Upgrade to latest xeus 3.2
The current requirement of xeus in CMakeLists.txt is 3.0.5. This should be upgraded ot the lates 3 version.

anutosh491:
Fixes #94


mcbarton:
@anutosh491 If this is the new minimum version, then probably worth changing the yml files for mamba development to check the ci hasn't picked up a version that's less than it.

anutosh491:
Hi @mcbarton 

So I don't think we are dependent on xeus. Rather xeus is pulled by xeus-zmq so we're good if we have an upper bound on xeus-zmq (`< 2.0` ). But yeah for the emscripten build we can try that.

CMakeLists.txt
```diff
@@ -79,13 +79,13 @@ endif()
...
 
 set(xtl_REQUIRED_VERSION 0.7.0)
-set(xeus_REQUIRED_VERSION 3.0.5)
+set(xeus_REQUIRED_VERSION 3.2.0)
 
 if (NOT TARGET xtl)
     find_package(xtl ${xtl_REQUIRED_VERSION} REQUIRED)
 endif ()
 if (NOT TARGET xeus AND NOT TARGET xeus-static)
-    find_package(xeus ${xeus_REQUIRED_VERSION} REQUIRED)
```
vgvassilev, line 88: We should have a cmake range here. We should be able to build correctly xeus-cpp outside of the conda context. In that case we will need cmake to define the supported ranges, too.

anutosh491: Hey Vassil, not fully sure what you mean by a range here. Something like this ?
```
set(xeus_MINIMUM_VERSION 3.2.0)
set(xeus_MAXIMUM_VERSION 4.0.0)

if (NOT TARGET xtl)
    find_package(xtl ${xtl_REQUIRED_VERSION} REQUIRED)
endif ()
if (NOT TARGET xeus AND NOT TARGET xeus-static)
    find_package(xeus ${xeus_MINIMUM_VERSION} REQUIRED)
    if (xeus_VERSION VERSION_LESS xeus_MINIMUM_VERSION OR xeus_VERSION VERSION_GREATER xeus_MAXIMUM_VERSION)
        message(FATAL_ERROR "xeus version must be between ${xeus_MINIMUM_VERSION} and ${xeus_MAXIMUM_VERSION}")
    endif ()
endif ()
```

alexander-penev (approved):

CMakeLists.txt
```diff
@@ -79,13 +79,13 @@ endif()
...
+    find_package(xeus REQUIRED)
```
alexander-penev, line 88: Maybe it's really good here to check the version of the found package if it's >= xeus_REQUIRED_VERSION

anutosh491: Done.

CMakeLists.txt
vgvassilev: `find_package(xeus ${xeus_REQUIRED_VERSION} REQUIRED)` should probably stay because that tells the cmake that the version in `xeus_REQUIRED_VERSION` is the minimal required version.

anutosh491: Yeah I think that makes sense.

PR #110, opened 2024-05-21
Attempt to removed xtl dependency

### LINKED ISSUE #108: Drop the dependency on xtl
We use xtl only for `xtl::prefix_path`. Let's drop this compile-time dependency.

anutosh491:
Fixes #108

### DIFF
--- a/src/xinspect.hpp
+++ b/src/xinspect.hpp
@@ -15,8 +15,6 @@
 
 #include <pugixml.hpp>
 
-#include <xtl/xsystem.hpp>
-
 #include "xeus-cpp/xbuffer.hpp"
 #include "xeus-cpp/xpreamble.hpp"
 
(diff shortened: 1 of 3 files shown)


anutosh491:
Hmm wait, I think pertaining to this line
https://github.com/jupyter-xeus/xeus/blob/1bbef7208ed3d92a90eadce752d3b56b7636f920/xeusConfig.cmake.in#L25, we might not be able to drop the dependency of xtl from xeus-cpp ( The dependency chain might be xeus-cpp -> xeus-zmq -> xeus -> xtl)

vgvassilev:
Ah, that's a good point... :(

JohanMabille:
I think most of the features we use from the xtl are actually specific to xeus, meaning we could actually move or duplicate the code in xeus. I'm not 100% sure though.

PR #113, opened 2024-05-22
Removing dirent dependency

anutosh491:
I don't think we use `dirent` for windows build anymore and can be removed.


mcbarton (commented):
@anutosh491 Unless we are expecting people to install versions 0.0.1-0.3.0 then its probably worth removing these from the readme. Also I don't know at which point the direct requirement for Windows builds was removed, so another reason to remove the old versions from the table.

anutosh491:
Ahh okay got you. So only `0.0.1` uses dirent and dirent was removed through this [commit](https://github.com/compiler-research/xeus-cpp/pull/11/commits/22af8f3d55014107adf35a982bfd6e9155956caf) dated back to July 22, 2023 and `0.1.0` was out by Nov, 2023. So I think we can get rid of the first patch version.

PR #115, opened 2024-05-22
Remove resource and include directories from kernel json

mcbarton:
When creating the interpreter we already detect and add the arguments for the resource and include directories here
https://github.com/mcbarton/xeus-cpp/blob/b58b15b83b27fe2e73c00c469346eea65301464d/src/xinterpreter.cpp#L37C1-L56C2 . This makes their addition in the kernel json file redundant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,15 +134,11 @@ endif ()
 function(configure_kernel kernel)
   set(XEUS_CPP_PATH "$ENV{PATH}")
   set(XEUS_CPP_LD_LIBRARY_PATH "$ENV{LD_LIBRARY_PATH}")
-  set(XEUS_CPP_RESOURCE_DIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/clang/${CPPINTEROP_LLVM_VERSION_MAJOR})
-  set(XEUS_CPP_INCLUDE_DIR ${CMAKE_INSTALL_PREFIX}/include)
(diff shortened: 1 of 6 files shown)


mcbarton:
@anutosh491 @vgvassilev This PR is ready for review.

vgvassilev:
> When creating the interpreter we already detect and add the arguments for the resource and include directories using CppInterOp here
https://github.com/mcbarton/xeus-cpp/blob/b58b15b83b27fe2e73c00c469346eea65301464d/src/xinterpreter.cpp#L37C1-L56C2 

Unfortunately that works on some platforms depending if we have installed compiler and we can find the compiler...

The resource-dir is particularly tricky to find and in CppInterOp we need to figure out how to make it part of the library. Only then we can drop it and be confident that we are not shooting our feet...

mcbarton:
> > When creating the interpreter we already detect and add the arguments for the resource and include directories using CppInterOp here
> > https://github.com/mcbarton/xeus-cpp/blob/b58b15b83b27fe2e73c00c469346eea65301464d/src/xinterpreter.cpp#L37C1-L56C2
> 
> Unfortunately that works on some platforms depending if we have installed compiler and we can find the compiler...
> 
> The resource-dir is particularly tricky to find and in CppInterOp we need to figure out how to make it part of the library. Only then we can drop it and be confident that we are not shooting our feet...

I'll convert this PR to a draft for now then, until that time comes where we are confident we can drop it.

PR #121, opened 2024-05-23
Update example notebook

anutosh491:
As we plan to deprecate xeus-cling once we release `1.0.0`, I think it makes sense to update our example notebook with most of what xeus-cling's [example notebook](https://github.com/jupyter-xeus/xeus-cling/blob/main/notebooks/xcpp.ipynb) supports.

I think we have most of what is required except Magics and Xtensor support.


vgvassilev:
@anutosh491, @mcbarton, @tharun571, is there a way to run these demos as part of the tests? This way we are guaranteed they will work.

mcbarton:
> @anutosh491, @mcbarton, @tharun571, is there a way to run these demos as part of the tests? This way we are guaranteed they will work.

Some parts of it are already being run as part of the tests in Jupyter Kernel Test, so should be possible. It won't work on Windows until the reason why XCppTests has been determined.

anutosh491:
> Some parts of it are already being run as part of the tests in Jupyter Kernel Test, so should be possible

Yes I haven't used the jupyter-kernel-test library personally (https://github.com/jupyter/jupyter_kernel_test) but I guess that is used to test out kernels.

vgvassilev:
In my very naive imagination we can/should have some binary that takes a notebook as parameter and returns some jsons for the results... If we don't have such a thing probably that's easy to implement. I think such a thing is critical to make sure xeus-cpp works consistently over time...

anutosh491:
Yeah that makes sense. I have the following thoughts
1) Let me open an issue to indicate introducing of a testing framework as the one you're talking about (takes a notebook as input and maybe returns some json as result)
2) That being said we might not want to block this PR as we didn't have most features like the documentation lookup through the inspect request etc when the notebook was added by Sylvain last year. I like the idea of the general audience seeing our updated notebook.
3) We might want to look at the testing framework not from a xeus-cpp level but from the xeus-stack (including all kernels). Just like the jupyter kernel test can test all kernels, the binary we're talking about might cover all of xeus-stack.

cc @vgvassilev @mcbarton  if you're thoughts match with mine, you can maybe review and merge and I'll address the 1st point then.

vgvassilev:
Ok, let’s open an issue for this and move forward with the pr.

anutosh491:
Raised an issue here #https://github.com/compiler-research/xeus-cpp/issues/122

PR #124, opened 2024-05-26
Add test for xinterpreter

tharun571:


tharun571:
@alexander-penev review please.

test/test_interpreter.cpp
```diff
@@ -645,4 +645,33 @@ TEST_SUITE("xutils_handler"){
...
+#endif
+
+TEST_SUITE("complete_request")
+{
+    TEST_CASE("completion_test")
+    {
+        std::vector<const char*> Args = {/*"-v", "resource-dir", "....."*/};
+        xcpp::interpreter interpreter((int)Args.size(), Args.data());
+
+        std::string code = "foo = a.isal";
```
alexander-penev, line 657: For autocomplete to work, we must first execute some code to include the appropriate header file and then attempt to use those functions.
Functions `isalnum` and `isalpha` are defined in `ctype.h`
Тherefore, before the complete function, we must execute this code through the interpreter: `#include <ctype.h>`
Something like:
`nl::json result = interpreter.execute_request("#include <ctype.h>", /*silent=*/false, ... etc );`

However, my experiments showed that maybe there is some problem in the implementation of the "complete" function in CppInterOp, and indeed even with the included library and the code that is appropriate, for example `auto i = isal` function does not return any suggestions. It is good to add an issue to report the CppInterOp bug.

It's a good idea to try if the test case code works in Jupyter Notebook or Jupyter Console.

tharun571: I tried it out in all the platforms and it seems that the code completion is not working for the functions involving importing from the header files. The updated code in this PR shows that it works for normal cases.

test/test_interpreter.cpp
```diff
@@ -645,4 +645,33 @@ TEST_SUITE("xutils_handler"){
...
+        REQUIRE(result["cursor_end"] == 2);
+        REQUIRE(result["status"] == "ok");
+        bool found = false;
+        for(auto& r : result["matches"]) {
+            if(r == "static" || r == "struct") {
+                found = true;
+                break;
+            }
+        }
+        REQUIRE(found == true);
```
alexander-penev, line 675: As written, everything will be OK if the function returns at least one of the expected results. I'm not sure this is a good approach for this test.
It is easiest to check if the field converted to a string is equal to {"struct", "strict"}, but that way there is a danger that in a newer version of the library (iostream) we will have more possible values, for example "stone". Then of course we will have to adjust the test.
Alternatively, we can check that the function returns everything expected without requiring that there are no other values, like so:
```suggestion
        size_t found = 0;
        for (auto& r : result["matches"]) {
            if (r == "static" || r == "struct") {
                found++;
            }
        }
        REQUIRE(found == 2);
```

tharun571: Makes sense.

alexander-penev (approved):

vgvassilev:
I am wondering why this did not increase code coverage.

tharun571:
> I am wondering why this did not increase code coverage.

It said it has already been covered, but this is the only piece of code calling that function. And also it shows some code has not been covered even when it has been covered. Maybe codecov is not 100% accurate.

vgvassilev:
We seem to have that coverage here: https://github.com/compiler-research/xeus-cpp/blob/e4a86ec9c8b1da841b448179203e3d0727474f30/test/test_xcpp_kernel.py#L25

It is good to have it on the other place, too. This explains why we did not see coverage increase.

Issue #54, opened 2024-03-22
Remove the PREFIX from generated binaries


SylvainCorlay:
We pass the install prefix as a compile time definition for the tag files, making relocation less easy (via the `XCPP_TAGFILES_DIR` compile time definition).

We should rather use `xtl::prefix_path` at runtime to prevent the need for binary relocation.

This issue also applies to xeus-cling.

vgvassilev:
I was thinking if we could rely on the environment here somehow. That way we can have one less dependency.

JohanMabille:
this function has been moved to `xeus`, so we should be able to use it without reintroducing the xtl dependency.

PR #127, opened 2024-05-29
Upgrade to xeus 5

### LINKED ISSUE #108: Drop the dependency on xtl

anutosh491:
Fixes #108 
Closes #110


mcbarton:
@anutosh491 maybe link this to the issues about removing the xtl dependency and update the readme about depenencies on the main

alexander-penev (approved):
Yes, everything looks very good, only the test crashes because of the removed xtl. If the test works out to not use the now removed xtl, the PR can be merged.

alexander-penev (changes requested):
Yes, everything looks very good, only the test crashes because of the removed xtl. If the test works out to not use the now removed xtl, the PR can be merged.

anutosh491:
> as it makes use of xtl when it loads the xtl header

Ahh I overlooked this. Thanks for the debug :)

mcbarton:
@anutosh491 will this also fix https://github.com/compiler-research/xeus-cpp/issues/54 since its related to xtl?

anutosh491:
Hmm, let me think through it. I think it might just solve that !

anutosh491:
> @anutosh491 will this also fix #54 since its related to xtl?

Well the 2 problems have some overlap but are a bit different. I think I have a way to tackle the issue, will solve it through a subsequent PR.

mcbarton:
@anutosh491 xeus_REQUIRED_VERSION needs to be version 5.1.0 now

anutosh491:
> @anutosh491 xeus_REQUIRED_VERSION needs to be version 5.1.0 now

Yeah I think I would use `xeus >= 5.0.0` in the environment. Just was debugging if we were able to fetch `5.1.0`. So required version could be `5.0.0`

environment-dev.yml
```diff
@@ -7,17 +7,17 @@ dependencies:
...
   - cppzmq
-  - xtl
   - CppInterOp>=1.3.0
   - pugixml
   - cpp-argparse <3.1
   - zlib
   # Test dependencies
   - pytest
-  - jupyter_kernel_test>=0.4.3
+  - jupyter_kernel_test>=0.5,<0.6
```
mcbarton, line 20: @anutosh491 Why is the newest version of jupyter_kernel_test (0.7.0) incompatible? I can look into making it compatible once I know why.

anutosh491: Ahh no specific reason for this change. Whenever we update the xeus-stack, we update all the kernels to the same versions for dependencies mostly (mainly looking at xeus-python) . For now we've decided to go with this version of jupyter_kernel_test for consistency. I am not sure if 0.7.0 is incompatible but for the first release of xeus-cpp (1.0.0) let's use this version of jupyter_kernel_test as it will be used across all xeus stack !

notebooks/xeus-cpp.ipynb
```diff
@@ -499,7 +497,7 @@
     "    nl::json mime_bundle_repr(const image& i)\n",
     "    {\n",
     "        auto bundle = nl::json::object();\n",
-    "        bundle[\"image/png\"] = xtl::base64encode(i.m_buffer.str());\n",
+    "        bundle[\"image/png\"] = xeus::base64encode(i.m_buffer.str());\n",
```
anutosh491, line 500: Well I thought this would do the job for us :(

mcbarton: @anutosh491 Don't you need to load a xeus header to use it?

anutosh491:
cc @alexander-penev @vgvassilev @mcbarton 
This is ready now !

anutosh491:
Ahh okay 1 thing. I'll address the readme too :)

src/xinterpreter.cpp
```diff
@@ -248,7 +248,8 @@ __get_cxx_version ()
             kernel_res["payload"] = nl::json::array();
             kernel_res["user_expressions"] = nl::json::object();
         }
-        return kernel_res;
+        cb(kernel_res);
+        return;
```
JohanMabille, line 252: no need for return here.

src/xinterpreter.cpp
```diff
@@ -237,7 +237,7 @@ __get_cxx_version ()
             /*
                 // Publish a mime bundle for the last return value if
                 // the semicolon was omitted.
-                if (!silent && output.hasValue() && trim(code).back() != ';')
+                if (!config.silent && output.hasValue() && trim(code).back() != ';')
```
JohanMabille, line 240: Should we uncomment this block or drop it? (I know it is not related to this PR, but I don't remember why we have it).

JohanMabille (approved):

alexander-penev (approved):

PR #130, opened 2024-05-31
Update pytest instructions in documentation

mcbarton:
This PR updates the pytest instructions in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ cd test
 ```
 and
 ```bash
+cd ../../test
 pytest -sv test_xcpp_kernel.py
 ```
 to perform the python tests.
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ cd test
 ```
 and
 ```bash
-pytest -sv .
+pytest -sv test_xcpp_kernel.py
 ```
 to perform the python tests.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -52,7 +52,7 @@ cd test
 ```
 and
 ```bash
-pytest -sv .
+pytest -sv test_xcpp_kernel.py
 ```
```
anutosh491, line 56: Ahh wait I am confused. So we create build and technically after running make install we have a test dir inside build. Do we also exoect the `test_xcpp_kernel.py` to be present there.

mcbarton: @anutosh491 It is on Linux and MacOS, as can be seen from the ci. Its not true for Windows (I found this out fixing the ci for Jupyter Kernel Test on Windows). I plan to fix this when I add instructions for building on Window

anutosh491: I am using a linux based device and I don't find this inside my build/test dir. will check again !

tharun571: There is no `test_xcpp_kernel.py` inside build/test. The way I think to check this test is to run this command inside the test dir with all the test files.

anutosh491: +1 Atleast the current cmake instruction don't move the `test_xcpp_kernel.py` file into `build/test` for me

mcbarton:
@anutosh491 I've updated the instructions.Are you happy with them now?

anutosh491 (approved):
LGTM

PR #133, opened 2024-05-31
Add xeus_zmq_REQUIRED_VERSION cmake

mcbarton:
We current set the required version in cmake by a variable for other packages but not xeus-zmq. This PR fixes this, and updates the documentation to make clear 1.0.2 is needed as a minimum and not 1.0.0


anutosh491:
I have two things here 
1) Let's wait for the update to xeus 5 PR to go in (will do that by today) and then wait for this PR to build on top of that.
2) Also we need to think if a minimum xeus_zmq version is even required. Cause if we're setting a minimum version for xeus .. that by itself might end up being sufficient. For eg xeus 5.0.0 is only comptabile with xeus-zmq 3.0.0 and xeus 4.0.0 (and above) is only comptabile with xeus-zmq 2.0.0 ... so technically ensuring a lower bound of xeus does ensure a lower bound on xeus-zmq

mcbarton:
> I have two things here
> 
>     1. Let's wait for the update to xeus 5 PR to go in (will do that by today) and then wait for this PR to build on top of that.
> 
>     2. Also we need to think if a minimum xeus_zmq version is even required. Cause if we're setting a minimum version for xeus .. that by itself might end up being sufficient. For eg xeus 5.0.0 is only comptabile with xeus-zmq 3.0.0 and xeus 4.0.0 (and above) is only comptabile with xeus-zmq 2.0.0 ... so technically ensuring a lower bound of xeus does ensure a lower bound on xeus-zmq

If xeus-zmq minimum version is not required i'll just drop the 1.0.2 from the cmake and remove the xeus-zmq in the readme.

vgvassilev (approved):
LGTM!

PR #134, opened 2024-05-31
Update clang format spec to c++17 standard

mcbarton:
We have the C++ standard in cmake set to c++17 but in the clang format spec its set to c++14 for some reason. This PR just updates the spec to the correct one.


anutosh491 (approved):
Hmm, this does make sense to me though I haven't looked into any of the previous clang format work/PRs before so let's wait for another approval.

mcbarton:
@vgvassilev can you take a look at this change? @anutosh491 would like someone with experience in clang format to review also.

PR #138, opened 2024-05-31
Add tests for xmagics

tharun571:


test/test_interpreter.cpp
```diff
@@ -626,6 +642,51 @@ TEST_SUITE("xmagics_apply"){
...
+
+        REQUIRE(kernel_res["status"] == "ok");
+    } 
+
+    TEST_CASE("cell magic with empty cell body") {
+        xcpp::xmagics_manager manager;
+
+        std::stringstream ss;
+        auto cerr_buff = std::cerr.rdbuf();
+        std::cerr.rdbuf(ss.rdbuf());
```
vgvassilev, line 681: I thought the doctest infrastructure had a way to capture the standard error and output streams but could not find anything in the repository or the documentation. @JohanMabille, do you happen to know?

tharun571: The Doctest testing framework does not have built-in support for capturing standard output or standard error streams. I had to manually redirect this stream and capture them.

vgvassilev: Ok, that's really unfortunate. Can you extract that logic into a RAII object so that future tests that require such a thing to reuse your code?

tharun571: Will add

test/test_interpreter.cpp
```diff
@@ -626,6 +662,49 @@ TEST_SUITE("xmagics_apply"){
...
+
+        preamble_manager["magics"].get_cast<xcpp::xmagics_manager>().apply("%magic1 qwerty", kernel_res);
+
+        REQUIRE(kernel_res["status"] == "ok");
+    } 
+
+    TEST_CASE("cell magic with empty cell body") {
+        xcpp::xmagics_manager manager;
+
+        std::stringstream ss;
```
vgvassilev, line 699: The stringstream should go in the StdErrRedirect object. I'd also like to rename it as `StreamRedirectRAII` which takes a parameter which stream to be redirected.

vgvassilev (approved):
LGMT!

test/test_interpreter.cpp
```diff
@@ -17,12 +17,52 @@
...
     #include <unistd.h>
 #endif
 
+/**
+ * @class StreamRedirectRAII
+ * @brief A RAII class to redirect a stream to a stringstream.
+ *
+ * This class redirects the output of a given std::ostream to a std::stringstream.
+ * The original stream is restored when the object is destroyed.
+ */
```
vgvassilev, line 34:
```suggestion
 /// A RAII class to redirect a stream to a stringstream.
 ///
 /// This class redirects the output of a given std::ostream to a std::stringstream.
 /// The original stream is restored when the object is destroyed.
```

vgvassilev: I've been used to this style of doxygen comments from llvm. Probably easier when building doxygen documentation one day in terms of configuration, etc. I did not find a lot of API documentation in the xeus-cpp, so maybe we should establish that style here...

test/test_interpreter.cpp
```diff
@@ -17,12 +17,52 @@
...
+    /**
+     * @brief Get the output that was written to the stream.
+     * @return A string containing the output that was written to the stream.
+     */
+    std::string getCaptured() {
+        return ss.str();
+    }
+
+private:
+    std::streambuf* old_stream_buff; ///< The original buffer of the stream.
```
vgvassilev, line 61:
```suggestion
    /// The original buffer of the stream.
    std::streambuf* old_stream_buff; 
```

Documenting the public api is generally enough as the expectation is that the private api will vary over time more and will more likely cause the documentation to go out of sync.

PR #142, opened 2024-06-08
Add notebook tests

### LINKED ISSUE #122: Introduce a testing framework to test out the example notebooks
As denoted in [this comment](https://github.com/compiler-research/xeus-cpp/pull/121#issuecomment-2127908892) we should have a testing framework (at the xeus-cpp level or at the overall xeus-stack level) that would accept a kernel ( Maybe different xeus kernels and their different possible versions) and a demo notebook and return some jsons for confirming the expected outputs.

This would ensure consistency for the kernels over the long run.

tharun571:
# Description

Using papermill library to execute the notebooks and check the necessary ouputs.

Tryng to fix #122 

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/test/test_xcpp_kernel.py
+++ b/test/test_xcpp_kernel.py
@@ -9,7 +9,8 @@
 import unittest
 import jupyter_kernel_test
 import platform
-
+import nbformat
+import papermill as pm
 
 class XCppCompleteTests(jupyter_kernel_test.KernelTests):
 
@@ -131,6 +132,52 @@ if platform.system() != 'Windows':
                 'mime': 'image/png'
             }
         ]
+
+    # Tests for Notebooks
+    class XCppNotebookTests(unittest.TestCase):
+
+        notebook_names = [
+            'xeus-cpp'
+            # Add more notebook names as needed
+        ]
+
+        def test_notebooks(self):
+            for name in self.notebook_names:
+
+                inp = f'Notebooks/{name}.ipynb'
+                out = f'Notebooks/{name}_output.ipynb'
+
+                with open(inp) as f:
+                    input_nb = nbformat.read(f, as_version=4)
+
+                try:
+                    # Execute the notebook
+                    executed_notebook = pm.execute_notebook( 
+                        inp, 
+                        out,
+                        log_output=True,
+                        kernel_name='xcpp20'
+                    )
+
+                    if executed_notebook is None:  # Explicit check for None or any other condition
+                        self.fail(f"Execution of notebook {name} returned None")
+                except Exception as e:
+                    self.fail(f"Notebook {name} failed to execute: {e}")
+
+                with open(out) as f:
+                    output_nb = nbformat.read(f, as_version=4)
+
+                check = True
+
+                # Iterate over the cells in the input and output notebooks
+                for i, (input_cell, output_cell) in enumerate(zip(input_nb.cells, output_nb.cells)):
+                    if input_cell.cell_type == 'code' and output_cell.cell_type == 'code':
+                        # If one cell has output and the other doesn't, set check to False
+                        if bool(input_cell.outputs) != bool(output_cell.outputs):
+                            self.fail(f"Cell {i} in notebook {name} has mismatched output presence")
+                        else:
+                            if input_cell.outputs != output_cell.outputs:
+                                self.fail(f"{input_output.get('text')} != {output_output.get('text')} Cell {i} in notebook {name} has mismatched output type")
 
 
 class XCppTests2(jupyter_kernel_test.KernelTests):
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -139,7 +139,7 @@ jobs:
         shell: bash -l {0}
         run: |
           cd test
-          pytest -sv .  --reruns 5
+          pytest -sv .
   
       - name: Python tests Windows Systems
         if: ${{ runner.os == 'windows' }}
--- a/docs/source/debug.rst
+++ b/docs/source/debug.rst
@@ -8,6 +8,7 @@ These steps are performed using the GNU Debugger (GDB), so ensure it is installe
 
 .. code-block:: bash
 
+    cd build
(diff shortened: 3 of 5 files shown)


test/test_xcpp_kernel.py
```diff
@@ -133,6 +135,24 @@ class XCppTests(jupyter_kernel_test.KernelTests):
...
+    )
+
+    # Read the executed notebook
+    with open('Notebooks/Untitled_output.ipynb') as f:
+        nb = nbformat.read(f, as_version=4)
+
+    # Get the output of the third cell
+    output = nb.cells[2].outputs[0]['text']
+
+    assert output == '12'
```
vgvassilev, line 154: Where does that 12 come from?

tharun571: In the notebook I had `cout<<12;` So I'm checking here if the 12 has actually been printed.

vgvassilev: We should take it from the notebook itself rather than hardcoding it.

tharun571: Yes, I take it from the notebook and check if the value is true. Line 148 to 152 is reading the notebook.

vgvassilev: I understand that but we should read the value of 12 from there too.

vgvassilev: I guess I am confused. Do we run the notebook somewhere?

test/test_xcpp_kernel.py
```diff
@@ -135,23 +134,36 @@ class XCppTests(jupyter_kernel_test.KernelTests):
...
-    with open('Notebooks/Untitled_output.ipynb') as f:
-        nb = nbformat.read(f, as_version=4)
-
-    # Get the output of the third cell
-    output = nb.cells[2].outputs[0]['text']
-
-    assert output == '12'
+    class XCppNotebookTests(unittest.TestCase):
+
+        inp = 'Notebooks/Untitled.ipynb'
```
vgvassilev, line 139: Now we need to make this a predefined list and have a test for each entry of that list. There should be some way to add programmatically new tests.

test/test_xcpp_kernel.py
```diff
@@ -132,6 +133,43 @@ class XCppTests(jupyter_kernel_test.KernelTests):
...
+                    if input_output.get('text') != output_output.get('text'):
+                        check = False
+                        break
+        
+        return check
+
+    # Tests for Notebooks
+    class XCppNotebookTests(unittest.TestCase):
+
+        def test_simple_notebook(self):
```
vgvassilev, line 169: I thought we can generate that test function, too...

test/test_xcpp_kernel.py
```diff
@@ -132,6 +133,43 @@ class XCppTests(jupyter_kernel_test.KernelTests):
...
+            self.assertEqual(execute_notebook('Notebooks/Untitled.ipynb', 'Notebooks/Untitled_output.ipynb'), True)
```
vgvassilev, line 170: What happens if you change the input expected values, will this break?

test/test_xcpp_kernel.py
tharun571: Just like the other classes, if the user wants to check a particular notebook he will add a definition with appropriate name. If the test fails somewhere this will help us know where it did. Let me know if you want me to make this into a list and iterate it instead.

test/test_xcpp_kernel.py
tharun571: Yes.

vgvassilev: Can you make that change and see if the CI fails?

test/test_xcpp_kernel.py
vgvassilev: Yes, it should be easy to add a new notebook and having a list is what we want I think.

test/test_xcpp_kernel.py
```diff
@@ -132,6 +133,54 @@ class XCppTests(jupyter_kernel_test.KernelTests):
...
+                            check = False
+                            break
+        
+        return check
+
+    # Tests for Notebooks
+    class XCppNotebookTests(unittest.TestCase):
+
+        notebook_pairs = [
+        ('Notebooks/test.ipynb', 'Notebooks/test_output.ipynb'),
```
vgvassilev, line 174: We do not need to force the user to write `test_output.ipynb` as this is an implementation detail on how we process the notebook. We can copy with a fixed prefix.

test/test_xcpp_kernel.py
vgvassilev: Ping.

test/test_xcpp_kernel.py
```diff
@@ -132,6 +133,52 @@ class XCppTests(jupyter_kernel_test.KernelTests):
...
+                            check = False
+                            break
+        
+        return check
+
+    # Tests for Notebooks
+    class XCppNotebookTests(unittest.TestCase):
+
+        notebook_names = [
+            'test',
```
vgvassilev, line 174: This notebook should be called something like `BasicFeatures` in which we record the basic features that xeus-cpp should support.

test/test_xcpp_kernel.py
vgvassilev: I still would like to introduce a different expected result in the test notebook and see the CI of this PR fail first.

test/Notebooks/test.ipynb
```diff
@@ -79,15 +79,7 @@
      "languageId": "c++"
     }
    },
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "12"
```
vgvassilev: I had in mind here to just put `13`, for instance. Can you do this?

vgvassilev:
Is this pull request still struggling to show a good user message if a notebook execution ends up with unexpected results or we are seeing something else?

vgvassilev (approved):
This is simply awesome! LGTM!

@JohanMabille, could we move that feature up the jupyter stack?

PR #143, opened 2024-06-11
Added functions to retrieve tag files/configs

### LINKED ISSUE #54: Remove the PREFIX from generated binaries

anutosh491:
# Description

We should be dependent on compile time definitions for fetching tagfiles and we should be using prefix_path from xeus to do the following at runtime as mentioned in the issue. Hence I've added relevant functions in the xutils.hpp file for the same.

Fixes #54 

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,9 @@ set(XCPP_TAGCONFS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/etc/xeus-cpp/tags.d)
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/include/xeus-cpp/xeus_cpp_config.hpp.in" 
                 "${CMAKE_CURRENT_SOURCE_DIR}/include/xeus-cpp/xeus_cpp_config.hpp")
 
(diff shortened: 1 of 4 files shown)


JohanMabille (approved):
LGTM. Next setp will be to restore the old config file, i.e. not gernerating it from cmake, but read the version from it in the CMakeLists.txt instead. This can be done in a dedicated PR.

anutosh491:
Thanks @JohanMabille for the review. Let's go ahead and merge this. I'll try to quickly address the above TODO in a dedicated PR :)

vgvassilev (approved):
LGTM!

SylvainCorlay:
This PR would probably be worth porting to xeus-cling.

anutosh491:
Sure I can look into it !

PR #149, opened 2024-06-18
Refactored register_preamble

anutosh491:
# Description

Addressed the [FIXME](https://github.com/compiler-research/xeus-cpp/blob/f071df023cb671df101b7176245acd8e97059806/src/xinterpreter.cpp#L394) comment in the code base. 

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/test/test_interpreter.cpp
+++ b/test/test_interpreter.cpp
@@ -407,7 +407,7 @@ TEST_SUITE("clone_magics_manager")
     {
         xcpp::xmagics_manager manager;
 
-        xcpp::xpreamble* clone = manager.clone();
+        std::unique_ptr<xcpp::xpreamble> clone = manager.clone();
 
         REQUIRE(clone != nullptr);
     }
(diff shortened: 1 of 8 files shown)


anutosh491:
@vgvassilev not sure if the above clang tidy suggestions should be applied. If we could use nodiscard but we can also do without it I suppose.

test/test_interpreter.cpp
```diff
@@ -651,9 +650,9 @@ TEST_SUITE("xsystem_clone")
     {
         xcpp::xsystem system;
 
-        xcpp::xpreamble* clone = system.clone();
+        std::unique_ptr<xcpp::xpreamble> clone = system.clone();
 
-        REQUIRE(dynamic_cast<xcpp::xsystem*>(clone) != nullptr);
+        REQUIRE(dynamic_cast<xcpp::xsystem*>(clone.get()) != nullptr);
```
vgvassilev, line 655: Does `xsystem` have multiple inheritance? If not we do not need `dynamic_cast`s here.

anutosh491: Addressed.

vgvassilev (approved):
LGTM!

PR #150, opened 2024-06-18
Config

anutosh491:
# Description

Addressing this [comment](https://github.com/compiler-research/xeus-cpp/pull/143#pullrequestreview-2124823990) raised by Johan

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,9 +30,6 @@ set(XCPP_TAGFILES_DIR ${CMAKE_CURRENT_SOURCE_DIR}/share/xeus-cpp/tagfiles)
 set(XEUS_CPP_CONF_DIR "etc/xeus-cpp")
 set(XCPP_TAGCONFS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/etc/xeus-cpp/tags.d)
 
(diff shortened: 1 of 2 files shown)


JohanMabille (changes requested):
I think we want to keep the `XEUS_CPP_VERSION_LABEL` token, and read it from the config file like the other ones.

vgvassilev:
I am lost. Why do we undo the work being done in this area?

JohanMabille:
As said in the [original PR](https://github.com/compiler-research/xeus-cpp/pull/128#pullrequestreview-2093024149), the building tool should not inject anything in the source code, this makes the code depending on the building tool. The original PR was merged to get the inspect feature work until we find a better solution. Now that we have one, we don't need the build tool to inject anything in the source code anymore and we can revert to the original behavior.

vgvassilev:
> As said in the [original PR](https://github.com/compiler-research/xeus-cpp/pull/128#pullrequestreview-2093024149), the building tool should not inject anything in the source code, this makes the code depending on the building tool. The original PR was merged to get the inspect feature work until we find a better solution. Now that we have one, we don't need the build tool to inject anything in the source code anymore and we can revert to the original behavior.

Ok, good. I misread the pull request on my phone. Apologies. LGTM!

JohanMabille (approved):

Issue #4, opened 2023-04-05
Implement a plugin system


vgvassilev:
Xeus-cpp should be able to attach clang-repl and other plugins such as clad or cppyy. We should investigate and implement plugin-system which is friendly to package managers.

@SylvainCorlay, you had a concrete proposal for directory structure that we wanted to adopt. Could you spell it out here?

cc: @alexander-penev, @grimmmyshini

SylvainCorlay:
The way I would go about it is the following:

Each clang-repl plugin provides a shared object (dynamic library) installed in the standard location (`$PREFIX/lib` most typically) providing a standardized API, for example a `load(xcpp::interpreter&)` free function to be invoked upon starting the plugin.

Each plugin gets installed with a "manifest" in `$PREFIX/share/xeus-cpp/plugins.d` in the form of a JSON file providing information about the plugin (the name of the dynamic library to be loaded, and metadata such as a description).

```cpp
const char* name = plugins["clad"];      // get the library to be loaded for the clad plugins 
void* handle = dlopen(name);             // load the library
if (handle != NULL)
{
    func* setup = dlsym(handle, "load");
    if (load != NULL)
    {
        int ret = setup(interpreter);     // setup the plugin, passing e.g. the interpreter instance to the setup function.
        return 0;
    }
}
return 1;
```

This can obviously be a lot more advanced. Plugins can provide more functions, be listed by listing the items in the `plugins.d` directory. Installing a plugin is done by placing files in the installation prefix, so that it can easily be done via a package manager.

SylvainCorlay:
Hey @vgvassilev. The timing is almost perfect. @derthorsten just pushed out the first replease of [xplugin](https://github.com/QuantStack/xplugin/).

PR #151, opened 2024-06-20
xassist implementation

tharun571:
# Description

- Introduced the "xassist" magic command to interact with large language models (LLMs) such as OpenAI and Gemini.
- Integrated the curl library to facilitate HTTP requests to the LLM APIs.
- Implemented functionality to save and load API keys for different models.
- Added support for maintaining context and chat history within the "xassist" command.
- Included necessary documentation to guide users on how to use the new features.
- Added tests to ensure the functionality works as expected and to help users verify their setup.

## Type of change

- [ ] Bug fix
- [x] New feature


SylvainCorlay:
I think this shows we need an extension system for magics, so that we don't keep to much in core xeus-cpp and have such features in extensions.

vgvassilev:
Perhaps we need to implement https://github.com/compiler-research/xeus-cpp/issues/4. @tharun571, can you take a look?

I guess here the question is should we accept the PR and move it as a plugin or wait for the plugin system?

JohanMabille:
I think we can accept the PR and move it to the plugin system when we have it, because the implementation does not spread across the whole kernel. My only concern is the dependency on curl, but we can probably hide it behind a building flag if we package a new version of xeus-cpp before the plugin system is available.

JohanMabille (approved):

CMakeLists.txt
```diff
@@ -203,6 +203,8 @@ set(XEUS_CPP_SRC
     src/xoptions.cpp
     src/xparser.cpp
     src/xutils.cpp
+    src/xmagics/xassist.hpp
```
vgvassilev, line 206: We should make these conditional to #ifdef __EMSCRIPTEN__ in cmake. Also
We should split the header file and the source file in the previous list.

vgvassilev (approved):
LGTM, the failures are actually regression in the master (see #155)

anutosh491:
I guess this would involve having a wasm build for curl (to support the esmcripten CI) ? 
I'm giving this an attempt today. Let me try to host a recipe for the same on emscripten-forge if we can get it to work that way.

anutosh491:
Raised a recipe here (https://github.com/emscripten-forge/recipes/pull/1332)

I guess having `libcurl-static` would suffice. We just need the `curl.h` file is it ?

tharun571:
> I guess having `libcurl-static` would suffice. We just need the `curl.h` file is it ?
Yes `curl.h` would help.

anutosh491:
I've added the recipe. Just needs to be reviewed by the folks there but in the meantime, any reviews from here (as to if we generate everything required or not would be great).

Currently I generate 
```
 │ │ Files in package:
 │ │   - bin/curl
 │ │   - bin/curl-config
 │ │   - include/curl/curl.h
 │ │   - include/curl/curlver.h
 │ │   - include/curl/easy.h
 │ │   - include/curl/header.h
 │ │   - include/curl/mprintf.h
 │ │   - include/curl/multi.h
 │ │   - include/curl/options.h
 │ │   - include/curl/stdcheaders.h
 │ │   - include/curl/system.h
 │ │   - include/curl/typecheck-gcc.h
 │ │   - include/curl/urlapi.h
 │ │   - include/curl/websockets.h
 │ │   - lib/libcurl.a
 │ │   - lib/pkgconfig/libcurl.pc
 │ │   - info/about.json
 │ │   - info/hash_input.json
 │ │   - info/index.json
 │ │   - info/licenses/COPYING
 │ │   - info/paths.json
 │ │   - info/recipe/build.sh
 │ │   - info/recipe/recipe.yaml
 │ │   - info/recipe/rendered_recipe.yaml
 │ │   - info/recipe/variant_config.yaml
 │ │   - info/tests/tests.yaml

```

PR #153, opened 2024-07-08
Refactoring some utilities

anutosh491:
Few utilities that we use are already implemented and tested through xeus. So we can just get rid of them !

### DIFF
--- a/src/xutils.cpp
+++ b/src/xutils.cpp
@@ -50,37 +50,6 @@ namespace xcpp
         exit(0);
     }
 
-    bool should_print_version(int argc, char* argv[])
-    {
-        for (int i = 0; i < argc; ++i)
-        {
-            if (std::string(argv[i]) == "--version")
-            {
-                return true;
-            }
-        }
-        return false;
-    }
-
-    std::string extract_filename(int &argc, char* argv[])
-    {
-        std::string res = "";
-        for (int i = 0; i < argc; ++i)
-        {
-            if ((std::string(argv[i]) == "-f") && (i + 1 < argc))
-            {
-                res = argv[i + 1];
-                for (int j = i; j < argc - 2; ++j)
-                {
-                    argv[j] = argv[j + 2];
-                }
-                argc -= 2;
-                break;
-            }
-        }
-        return res;
-    }
-
     interpreter_ptr build_interpreter(int argc, char** argv)
     {
         std::vector<const char*> interpreter_argv(argc);
--- a/include/xeus-cpp/xutils.hpp
+++ b/include/xeus-cpp/xutils.hpp
@@ -25,12 +25,6 @@ namespace xcpp
     void stop_handler(int sig);
 
     XEUS_CPP_API
-    bool should_print_version(int argc, char* argv[]);
-
-    XEUS_CPP_API
-    std::string extract_filename(int &argc, char* argv[]);
-
-    XEUS_CPP_API
     interpreter_ptr build_interpreter(int argc, char** argv);
 
     XEUS_CPP_API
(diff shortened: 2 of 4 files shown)


anutosh491:
We should also be able to used the [get_start_message ](https://github.com/jupyter-xeus/xeus/blob/a8a489214493adc42cdaa70f1460a87bf46546e7/src/xhelper.cpp#L12) function instead of the following 
https://github.com/compiler-research/xeus-cpp/blob/4745f724d3b2fdd674a1a46270f27972caf5984b/src/main.cpp#L109-L114

anutosh491:
Looked straightforward but for some reason the test_xcpp_kernel.py test fails for some reason. Need to have a look into it

anutosh491:
@vgvassilev @mcbarton I think this might have something to do with the patches in CppInterOp (y'all could help by confirming) cause through backtrace we get 
```
error: error reading '/Users/runner/micromamba/envs/xeus-cpp/include': Is a directory
1 error generated.
Error: signal 11:
0   libxeus-cpp.0.5.0dev.dylib          0x000000010d492fbe _ZN4xcpp7handlerEi + 46
1   libsystem_platform.dylib            0x00007ff809df75ed _sigtramp + 29
2   ???                                 0x0000000000000004 0x0 + 4
3   libclangCppInterOp.18.1.dylib       0x0000000117a94620 _ZN5clang17IncrementalParserC2ERNS_11InterpreterENSt3__110unique_ptrINS_16CompilerInstanceENS3_14default_deleteIS5_EEEERN4llvm11LLVMContextERNS9_5ErrorE + 208
4   libclangCppInterOp.18.1.dylib       0x0000000117a97cca _ZN5clang11InterpreterC2ENSt3__110unique_ptrINS_16CompilerInstanceENS1_14default_deleteIS3_EEEERN4llvm5ErrorE + 346
5   libclangCppInterOp.18.1.dylib       0x0000000117a97f44 _ZN5clang11Interpreter6createENSt3__110unique_ptrINS_16CompilerInstanceENS1_14default_deleteIS3_EEEE + 84
6   libclangCppInterOp.18.1.dylib       0x0000000117a78e42 _ZN6compat22createClangInterpreterERNSt3__16vectorIPKcNS0_9allocatorIS3_EEEE + 386
7   libclangCppInterOp.18.1.dylib       0x0000000117a78a73 _ZN3Cpp11InterpreterC2EiPKPKcS2_RKNSt3__16vectorINS5_10shared_ptrIN5clang19ModuleFileExtensionEEENS5_9allocatorISA_EEEEPvb + 883
8   libclangCppInterOp.18.1.dylib       0x0000000117a6d3c9 CreateInterpreter + 1833
9   libxeus-cpp.0.5.0dev.dylib          0x000000010d456c44 _Z17createInterpreterRKNSt3__16vectorIPKcNS_9allocatorIS2_EEEE + 1492
```

anutosh491:
Caught the bug here. Sadly I was looking at all places except the function I had changed (the stack trace didn't help and confused me even more). 

It's an error with how xeus deals with extract_filename function, so have to fix that upstream.

vgvassilev:
Why the nightly builds succeed then?

anutosh491:
This should be fixed after the next xeus release as https://github.com/jupyter-xeus/xeus/pull/409 has been merged now !

src/xutils.cpp
```diff
@@ -50,37 +50,6 @@ namespace xcpp
...
-                {
-                    argv[j] = argv[j + 2];
-                }
-                argc -= 2;
-                break;
-            }
-        }
-        return res;
-    }
-
```
vgvassilev, line 83: I'd keep the tests as upstream testing was a bit lacking last time I saw.

anutosh491: Ahhh no worries I made sure to add a test here (https://github.com/anutosh491/xeus/blob/13aea3049c99df566ae54061432b57b100747e91/test/test_unit_kernel.cpp#L56-L65)

PR #156, opened 2024-09-27
XAssist - add model freedom and ollama support

tharun571:
# Description
- Users can now select and use a specific model within their respective LLM.
- Users can set the URL for Ollama and can add a model and utilize the LLM.
- Updated the documentation to reflect the new features and changes.
- Added tests for the new Ollama support.

## Type of change

- [ ] Bug fix
- [x] New feature


anutosh491:
The failing CI (installing mamba/ micromamba) is due to the new release 
Pinning the version should fix it 
```
- uses: mamba-org/setup-micromamba@v1
  with:
    micromamba-version: '1.5.10-0'
```

Let's do this for now, untill the issue is fixed

src/xmagics/xassist.cpp
```diff
@@ -60,6 +60,80 @@ namespace xcpp
...
+            {
+                std::getline(in, url);
+                in.close();
+                return url;
+            }
+
+            std::cerr << "Failed to open file for reading URL for model " << model << std::endl;
+            return "";
+        }
+    };
```
anutosh491, line 135: Technically any utility function like these should be tested for (either in test_interpreter.cpp itself or maybe through a test_utils.cpp if possible) 

But that can be done through subsequent PRs (an issue can be opened to track the same) !

anutosh491 (approved):
Clang-tidy reviews should be addressed. Apart from that it LGTM.

I would just wait for the green flag by @vgvassilev before merging !

vgvassilev (approved):
LGTM, let's ignore the infrastructure issue of the clang-tidy bot.

PR #161, opened 2024-10-16
Refactor Xassist, Test Framework

tharun571:
# Description

- Added xassist method to handle special cases.
- Updated OpenAI call method.
- Removed unnecessary line in testing framework.

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/src/xmagics/xassist.cpp
+++ b/src/xmagics/xassist.cpp
@@ -19,7 +19,6 @@
 using json = nlohmann::json;
 
 // TODO: Implement xplugin to separate the magics from the main code.
-// TODO: Add support for open-source models.
 namespace xcpp
 {
     class api_key_manager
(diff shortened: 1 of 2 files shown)


anutosh491:
Just curious as to what exactly the refactor is doing here ? Maybe explain through a simple example ?

tharun571:
> Just curious as to what exactly the refactor is doing here ? Maybe explain through a simple example ?

I was going through whatever code I wrote and checked for possible bugs. This PR takes care of scenarios when the prompt has special characters or anything involving disrupting the JSON to call the LLM.

src/xmagics/xassist.cpp
```diff
@@ -279,6 +279,53 @@ namespace xcpp
...
+                    else
+                    {
+                        escaped += c;
+                    }
+                    break;
+            }
+        }
+        return escaped;
+    }
+
```
anutosh491, line 328: > anything involving disrupting the JSON to call the LLM.

Hmm, so should there be a test required here as to check here to check this claim ?

tharun571: > Hmm, so should there be a test required in this PR to check this claim ?

I did try but got in a predicament where `escape_special_cases("""{}") == "\"\"{}"` yields false but works in the kernel. So I tested them manually and pushed the code

vgvassilev (approved):
LGTM!

PR #164, opened 2024-10-17
Refactor CMakeLists.txt

anutosh491:
Removed redundancy and adds more consistency especially for the wasm build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,6 +212,9 @@ macro(xeus_cpp_set_common_options target_name)
     if (MSVC)
         target_compile_options(${target_name} PUBLIC /wd4251 /wd4141)
         target_compile_options(${target_name} PUBLIC /wd4018 /wd4267 /wd4715 /wd4146 /wd4129)
+        target_compile_options(${target_name} PUBLIC /EHsc)
+    else ()
+        target_compile_options(${target_name} PUBLIC -fexceptions)
     endif ()
 
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,15 +51,16 @@ message(STATUS "Building xeus-cpp v${${PROJECT_NAME}_VERSION}")
 # =============
 
 option(XEUS_CPP_BUILD_STATIC "Build xeus-cpp static library" ON)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -98,31 +100,6 @@ if (NOT TARGET xeus AND NOT TARGET xeus-static)
...
-    if(NOT EMSCRIPTEN)
-        add_compile_options(-Wunused-parameter -Wextra -Wreorder)
-    endif()
-endif ()
-
-if(EMSCRIPTEN)
-    # ENV (https://github.com/emscripten-core/emscripten/commit/6d9681ad04f60b41ef6345ab06c29bbc9eeb84e0)
-    set(EMSCRIPTEN_FEATURES "${EMSCRIPTEN_FEATURES} -s \"EXTRA_EXPORTED_RUNTIME_METHODS=[ENV']\"")
-endif()
-
```
anutosh491, line 125: These are redundant as are being addressed through `macro(xeus_cpp_set_common_options target_name)` so can be totally removed from here.

anutosh491:
@vgvassilev this should be ready (P.S i made the exact same change on xeus-r few days ago and I realized they would be applicable on xeus-cpp too so pretty confident :) )

CMakeLists.txt
JohanMabille: I think some flags are still missing in the macro (-fexceptions and /EHsc, I may have missed others).

anutosh491: Hmm, having a look !

anutosh491: Yupp those were the only ones. Should be ready now !

CMakeLists.txt
```diff
@@ -347,7 +332,6 @@ macro(xeus_cpp_create_target target_name linkage output_name)
         # Curl initialised specifically for xassist
         target_link_libraries(${target_name} PUBLIC ${XEUS_CPP_XEUS_TARGET} clangCppInterOp pugixml argparse::argparse curl)
     else ()
-        # TODO : Add curl support for emscripten
```
anutosh491, line 350: Removed TODO which can't be resolved as mentioned here (https://github.com/emscripten-forge/recipes/pull/1332#issuecomment-2371509339)

JohanMabille (approved):

PR #168, opened 2024-11-05
Fix logic for resource dir flag

anutosh491:
The logic for the resource dir flag looks corrupt. 

I am guessing we need to check if the flag is available, if not we need CppInterOp to fetch it for us. If flag is found add the arguments else return error.

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -37,12 +37,13 @@ using Args = std::vector<const char*>;
 
 void* createInterpreter(const Args &ExtraArgs = {}) {
   Args ClangArgs = {/*"-xc++"*/"-v"}; // ? {"-Xclang", "-emit-llvm-only", "-Xclang", "-diagnostic-log-file", "-Xclang", "-", "-xc++"};
-  if (std::find(ExtraArgs.begin(), ExtraArgs.end(), "-resource-dir") == ExtraArgs.end()) {
-    std::string resource_dir = Cpp::DetectResourceDir();
-    if (resource_dir.empty())
-      std::cerr << "Failed to detect the resource-dir\n";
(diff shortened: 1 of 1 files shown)


src/xinterpreter.cpp
```diff
@@ -37,12 +37,14 @@ using Args = std::vector<const char*>;
 
 void* createInterpreter(const Args &ExtraArgs = {}) {
   Args ClangArgs = {/*"-xc++"*/"-v"}; // ? {"-Xclang", "-emit-llvm-only", "-Xclang", "-diagnostic-log-file", "-Xclang", "-", "-xc++"};
-  if (std::find(ExtraArgs.begin(), ExtraArgs.end(), "-resource-dir") != ExtraArgs.end()) {
+  if (std::find(ExtraArgs.begin(), ExtraArgs.end(), "-resource-dir") == ExtraArgs.end()) {
```
vgvassilev, line 40: I'd keep only that change and the rest should not be changed.

src/xinterpreter.cpp
```diff
@@ -37,12 +37,15 @@ using Args = std::vector<const char*>;
...
-    ClangArgs.push_back(resource_dir.c_str());
+  if (std::find(ExtraArgs.begin(), ExtraArgs.end(), "-resource-dir") == ExtraArgs.end())
+  {
+      std::string resource_dir = Cpp::DetectResourceDir();
+      if (resource_dir.empty())
+      {
+        std::cerr << "Failed to detect the resource-dir\n";
+      }
+      ClangArgs.push_back("-resource-dir");
+      ClangArgs.push_back(resource_dir.c_str());
```
vgvassilev, line 48: Please make this patch one line as it should be.

anutosh491:
Closing as completed in #169

PR #169, opened 2024-11-05
Fix logic for resource dir flag

anutosh491:
The logic for the resource dir flag looks corrupt.

I am guessing we need to check if the flag is available, if not we need CppInterOp to fetch it for us. If flag is found add the arguments else return error.


vgvassilev (changes requested):
This should be really a change of 1 token. From `==` to `!=`.

PR #170, opened 2024-11-05
Stop fetching emsdk from conda-forge

anutosh491:
# Description

emsdk's recipe from conda-forge is no longer being maintained. Hence preferably we should clone it and install it accordingly.

## Type of change

- [ ] Bug fix
- [ ] New feature


mcbarton:
@anutosh491 why are we still using 3.1.45 when the latest version is 3.1.71?

PR #171, opened 2024-11-08
Fix finding Resource Dir

anutosh491:
# Description

`std::find `operates based on the o`perator==` of the type being compared. For `const char*`, the `operator== `compares the memory addresses (the pointer values), not the contents of the strings.

Hence although we should fetch resource dir from the kernel.json file, we end up entering the block checking for resource dir and trying getting it through CppInterOp. This should not be the case.

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -37,7 +37,11 @@ using Args = std::vector<const char*>;
 
 void* createInterpreter(const Args &ExtraArgs = {}) {
   Args ClangArgs = {/*"-xc++"*/"-v"}; // ? {"-Xclang", "-emit-llvm-only", "-Xclang", "-diagnostic-log-file", "-Xclang", "-", "-xc++"};
-  if (std::find(ExtraArgs.begin(), ExtraArgs.end(), "-resource-dir") == ExtraArgs.end()) {
+  auto it = std::find_if(ExtraArgs.begin(), ExtraArgs.end(), [](const char* arg) {
+    return std::strcmp(arg, "-resource-dir") == 0;
+  });
+
+  if (it == ExtraArgs.end()) {
     std::string resource_dir = Cpp::DetectResourceDir();
     if (resource_dir.empty())
       std::cerr << "Failed to detect the resource-dir\n";


vgvassilev (changes requested):
There is nothing wrong with the current approach:  https://godbolt.org/z/5GzTeYo93 I do not think this needs fixing. Consider closing this PR.

src/xinterpreter.cpp
```diff
@@ -37,7 +37,11 @@ using Args = std::vector<const char*>;
 
 void* createInterpreter(const Args &ExtraArgs = {}) {
   Args ClangArgs = {/*"-xc++"*/"-v"}; // ? {"-Xclang", "-emit-llvm-only", "-Xclang", "-diagnostic-log-file", "-Xclang", "-", "-xc++"};
-  if (std::find(ExtraArgs.begin(), ExtraArgs.end(), "-resource-dir") == ExtraArgs.end()) {
+  auto it = std::find_if(ExtraArgs.begin(), ExtraArgs.end(), [](const char* arg) {
+    return std::strcmp(arg, "-resource-dir") == 0;
+  });
+
+  if (it == ExtraArgs.end()) {
```
vgvassilev, line 44:
```suggestion
  if (std::find_if(ExtraArgs.begin(), ExtraArgs.end(), [](std::string s) {
    return s == "-resource-dir";}) == ExtraArgs.end()) {
```

vgvassilev (commented):
I think I see what you mean, because we have an `std::vector<const char*>` we can't really use the operator== of std::string. Apologies for the oversight.

src/xinterpreter.cpp
```diff
@@ -37,11 +37,8 @@ using Args = std::vector<const char*>;
 
 void* createInterpreter(const Args &ExtraArgs = {}) {
   Args ClangArgs = {/*"-xc++"*/"-v"}; // ? {"-Xclang", "-emit-llvm-only", "-Xclang", "-diagnostic-log-file", "-Xclang", "-", "-xc++"};
-  auto it = std::find_if(ExtraArgs.begin(), ExtraArgs.end(), [](const char* arg) {
-    return std::strcmp(arg, "-resource-dir") == 0;
-  });
-
-  if (it == ExtraArgs.end()) {
+  if (std::find_if(ExtraArgs.begin(), ExtraArgs.end(), [](std::string s) {
```
vgvassilev, line 40:
```suggestion
  if (std::find_if(ExtraArgs.begin(), ExtraArgs.end(), [](const std::string& s) {
```

PR #172, opened 2024-11-08
Address resource dir config for kernels

anutosh491:
# Description

See https://discord.com/channels/1235591463472074924/1235591601523396680/1304415164753379441

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,13 @@ find_package(pugixml REQUIRED)
 # Configuration
 # =============
 
+if (CPPINTEROP_LLVM_VERSION_MAJOR VERSION_GREATER ${CMAKE_CXX_COMPILER_VERSION})
+    message(WARNING "The resource directory being used corresponds to clang version ${CMAKE_CXX_COMPILER_VERSION}. \
+        It should be updated to match clang version ${CPPINTEROP_LLVM_VERSION_MAJOR} once cppinterop \
+        can provide a compatible resource directory.")
+endif()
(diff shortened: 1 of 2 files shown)


anutosh491:
My understanding here is this.

For providing the resource dir

1) We could either directly provide it through kernel.json or by manually while creating the interpreter.
2) If the above is not done, we would need to use `DetectResourceDir` from CppInterOp.

But as can be seen in DetectResourceDir we have this (https://github.com/compiler-research/CppInterOp/blob/826be787bb542fe2394a425de78613a4da3eae15/lib/Interpreter/CppInterOp.cpp#L2760)

So as our resource dir is being provided out of clang (based from cxx-compiler) ... the latest that can be provided is version `17` but we use cppinterop compatible with latest clang version (so `CLANG_VERSION_MAJOR_STRING` is 19) and we don't satisfy the above condition.

Hence the ways to approach this is 
1) Maybe not use cxx-compiler and introduce a dependency on clang binary (so that we can get access to latest resource dir)
2) Or fix `kernel.json` to reference the correct resource dir available out of cxx-compiler and provide the same for tests manually 
```
        std::vector<const char*> Args = {"-v", "-resource-dir", XEUS_CPP_RESOURCE_DIR};
        xcpp::interpreter interpreter((int)Args.size(), Args.data());  
```

anutosh491:
Not ready yet. Bad attempt to be fair. Converting to draft

anutosh491:
Closing as probably not the correct way to address the above https://discord.com/channels/1235591463472074924/1235591601523396680/1304415164753379441

PR #180, opened 2024-11-18
Remove xdemangle as it is redundant

### LINKED ISSUE #179: Is xdemangle.hpp redundant ?
I see that `xdemangle.hpp` provides the demangle function.

https://github.com/compiler-research/xeus-cpp/blob/e9df705c9f6a4e6b7f7f0b7655726ca4dc294ce7/src/xdemangle.hpp#L41-L42

This should technically be used in `xinspect` in the implementation of `find_type` (as done for [xeus-cling](https://github.com/jupyter-xeus/xeus-cling/blob/93d61b98518c494fc0e195844c392fc0d40dd0d2/src/xinspect.hpp#L128))

But we are currently using `find_type_slow` (the function name suggests this might not be the best approach ?!)

https://github.com/compiler-research/xeus-cpp/blob/e9df705c9f6a4e6b7f7f0b7655726ca4dc294ce7/src/xinspect.hpp#L82-L99

But if we are not trying to implement a "find_type_fast" and are comfortable with what we have, let's get rid of xdemangle.hpp ?

faze-geek:
# Description

Fixes #179 

## Type of change

- [ ] Bug fix
- [ ] New feature


anutosh491 (approved):
Looks good merging

Issue #176, opened 2024-11-15
Support all magics available in xeus-cling


anutosh491:
I currently see `xeus-cling` suporting few magics (https://xeus-cling.readthedocs.io/en/latest/magics.html)

- [ ] %%file
- [ ] %%timeit
- [ ] %%executable

I think %%file through os.hpp/os.cpp should work right off the bat, so that can be enabled.

anutosh491:
Just framed some dummy code for `%%timeit` (untested ... somebody could pick it up and test it)

https://gist.github.com/anutosh491/8ef8fddac545acfad05039164679ecf4

PR #181, opened 2024-11-18
Enable the `%%file` xmagic

faze-geek:
# Description

Towards #176

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,9 +182,7 @@ set(XEUS_CPP_SRC
 
 if(NOT EMSCRIPTEN)
     list(APPEND XEUS_CPP_SRC
-        #src/xmagics/xassist.hpp
         src/xmagics/xassist.cpp
-        #src/xmagics/os.hpp
         src/xmagics/os.cpp
     )
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,9 +182,9 @@ set(XEUS_CPP_SRC
 
 if(NOT EMSCRIPTEN)
     list(APPEND XEUS_CPP_SRC
-        src/xmagics/xassist.hpp
+        #src/xmagics/xassist.hpp
         src/xmagics/xassist.cpp
-        src/xmagics/os.hpp
+        #src/xmagics/os.hpp
         src/xmagics/os.cpp
     )
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,14 +169,12 @@ set(XEUS_CPP_HEADERS
     #src/xinspect.hpp
     #src/xsystem.hpp
     #src/xparser.hpp
-    #src/xmagics/os.hpp
 )
 
 set(XEUS_CPP_SRC
     src/xholder.cpp
     src/xinput.cpp
     src/xinterpreter.cpp
-    src/xmagics/os.cpp
     src/xoptions.cpp
     src/xparser.cpp
     src/xutils.cpp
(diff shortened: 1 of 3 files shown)


faze-geek:
Let me know if I need to add more tests. I wrote them for whatever functionalities were supported in the documentation.

anutosh491:
Please make the following changes 

1) In cmakelists.txt : Move oss header & source to the not emscripten case.
```
set(XEUS_CPP_HEADERS
    ....
    #src/xmagics/os.hpp
)

set(XEUS_CPP_SRC
    ....
    src/xinterpreter.cpp
    src/xmagics/os.cpp
    ....
)

if(NOT EMSCRIPTEN)
    list(APPEND XEUS_CPP_SRC
        src/xmagics/xassist.hpp
        src/xmagics/xassist.cpp
    )
endif()
```

2) in `src/xinterpreter.cpp` : Do the same as above.

```
#include "xmagics/os.hpp"
#ifndef EMSCRIPTEN
#include "xmagics/xassist.hpp"
#endif
```

anutosh491:
Apart from that. The tests look decent to me. Maybe @mcbarton could also review this for us !

vgvassilev:
@mcbarton, do we not have code coverage bots checking the test suite coverage for xeus-cpp? I remember we had that but maybe I am confusing myself...

mcbarton:
> @mcbarton, do we not have code coverage bots checking the test suite coverage for xeus-cpp? I remember we had that but maybe I am confusing myself...

Isn't the test coverage being shown in this message? https://github.com/compiler-research/xeus-cpp/pull/181#issuecomment-2482751686

CMakeLists.txt
```diff
@@ -186,6 +184,8 @@ if(NOT EMSCRIPTEN)
     list(APPEND XEUS_CPP_SRC
         src/xmagics/xassist.hpp
         src/xmagics/xassist.cpp
+        src/xmagics/os.hpp
```
vgvassilev, line 187: Do we need to list the local `hpp` files?

anutosh491: Guessing we can do without them. Trying it out.

anutosh491: The tests pass, so I think we can do without them

CMakeLists.txt
```diff
@@ -169,23 +169,23 @@ set(XEUS_CPP_HEADERS
...
     src/xutils.cpp
 )
 
 if(NOT EMSCRIPTEN)
     list(APPEND XEUS_CPP_SRC
-        src/xmagics/xassist.hpp
+        #src/xmagics/xassist.hpp
         src/xmagics/xassist.cpp
+        #src/xmagics/os.hpp
+        src/xmagics/os.cpp
```
vgvassilev, line 188: Can we remove the commented lines?

anutosh491: Done.

anutosh491:
I am guessing this is ready ?

vgvassilev:
The suggested by clang-format changes look good.

faze-geek:
> The suggested by clang-format changes look good.
Done !

anutosh491:
I see a duplication of tests here. We already had tests for the writefile operation.

https://github.com/compiler-research/xeus-cpp/blob/ca8c42e95b6b1aa5a84f31d6285af83e0de5cf04/test/test_interpreter.cpp#L527-L590

Can someone confirm this for me. The tests look similar though this looks better framed. Maybe @kr-2003 would like to help (maybe get rid of the first set of tests)

anutosh491:
> The only difference I can see is "removing the txt file after every subtest".

Not worth it. Let's just keep the better test

PR #183, opened 2024-11-19
Skip computing resource dir and include paths for xeus-cpp-lite

anutosh491:
While using xeus-cpp-lite ( or rather running clang-repl in the browser) , local paths to system includes or resource dirs don't make sense.

What should happen here is content out of 
1) Sysroot provided by emsdk (provided already)
2) Resource dir hopefully provided by CppInterOp (in the works)

Should be collected in one location and then we need to load it using emscripten's `--preload-file` ( we can use this flag only once as it creates a .data file which comprises of all the headers/file we need)

So 
1) The resource dir path present in the kernel.json is of no use
```
  "argv": [
      "/Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/bin/xcpp",
      "-f",
      "{connection_file}",
      "-resource-dir", "/Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/lib/clang/19",
      "-I", "/Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include",
      "-std=c++20"
  ],
```
2) Using `DetectResourceDir` and  `DetectSystemCompilerIncludePaths` can't be done as both depend on `exec` (marked as won't fix on emscripten)

3) `Init_includes()` adds `/Users/anutosh491/micromamba/envs/xeus-cpp-wasm-host/include` to the cc1 command but this doesn't make sense either.

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -28,6 +28,7 @@ using Args = std::vector<const char*>;
 
 void* createInterpreter(const Args &ExtraArgs = {}) {
   Args ClangArgs = {/*"-xc++"*/"-v"}; // ? {"-Xclang", "-emit-llvm-only", "-Xclang", "-diagnostic-log-file", "-Xclang", "-", "-xc++"};
+#ifndef EMSCRIPTEN
   if (std::find_if(ExtraArgs.begin(), ExtraArgs.end(), [](const std::string& s) {
     return s == "-resource-dir";}) == ExtraArgs.end()) {
     std::string resource_dir = Cpp::DetectResourceDir();
@@ -42,6 +43,7 @@ void* createInterpreter(const Args &ExtraArgs = {}) {
     ClangArgs.push_back("-isystem");
     ClangArgs.push_back(CxxInclude.c_str());
   }
+#endif
   ClangArgs.insert(ClangArgs.end(), ExtraArgs.begin(), ExtraArgs.end());
   // FIXME: We should process the kernel input options and conditionally pass
   // the gpu args here.
@@ -357,7 +359,9 @@ __get_cxx_version ()
 
     void interpreter::init_includes()
     {
+#ifndef EMSCRIPTEN
         Cpp::AddIncludePath((xeus::prefix_path() + "/include/").c_str());
+#endif
     }
 
     void interpreter::init_preamble()


anutosh491:
The preloaded file would end up in FS/include or FS/include/wasm32-emscripten which are the only two places from where we can access the headers.

src/xinterpreter.cpp
```diff
@@ -357,7 +359,9 @@ __get_cxx_version ()
 
     void interpreter::init_includes()
     {
+#ifndef EMSCRIPTEN
         Cpp::AddIncludePath((xeus::prefix_path() + "/include/").c_str());
```
vgvassilev, line 363: Does that hurt on emscripten? I'd think it's just some random non-existing folder?

anutosh491: Yeah random non existing folder that won't contribute to our solution :)

If you look at Anubhab's solution (https://wasmdemo.argentite.me/) or any other solutions that interest us (https://binji.github.io/wasm-clang/) we are only interested in two folders is what I can see 
1) include ( for `/include/c++/v1`)
2) include/wasm32-emscripten  (for `/include/wasm32-emscripten/c++/v1`)

```
// 1st link
"" -cc1 -triple wasm32-unknown-emscripten -emit-obj -mrelax-all -disable-free -clear-ast-before-backend -main-file-name "<<< inputs >>>" -mrelocation-model static -mframe-pointer=none -ffp-contract=on -fno-rounding-math -mconstructor-aliases -target-cpu generic -fvisibility=hidden -debugger-tuning=gdb -v -fcoverage-compilation-dir=/ -resource-dir lib/clang/17 -internal-isystem /include/wasm32-emscripten/c++/v1 -internal-isystem /include/c++/v1 -internal-isystem lib/clang/17/include -internal-isystem /include/wasm32-emscripten -internal-isystem /include -std=c++17 -fdeprecated-macro -fdebug-compilation-dir=/ -ferror-limit 19 -fgnuc-version=4.2.1 -fcxx-exceptions -fexceptions -fincremental-extensions -o "<<< inputs >>>.o" -x c++ "<<< inputs >>>"
clang -cc1 version 17.0.0 based upon LLVM 17.0.0git default target wasm32-unknown-emscripten
ignoring nonexistent directory "/include/wasm32-emscripten/c++/v1"
ignoring nonexistent directory "lib/clang/17/include"
ignoring nonexistent directory "/include/wasm32-emscripten"
/include/c++/v1
/include

// 2nd link
clang -cc1 -emit-obj -disable-free -isysroot / -internal-isystem /include/c++/v1 -internal-isystem /include -internal-isystem /lib/clang/8.0.1/include -ferror-limit 19 -fmessage-length 80 -fcolor-diagnostics -O2 -o test.o -x c++ test.cc
```

We eventually need to fetch anything we are interested in from one of the two folders and that is what --preload-file would do for us !

anutosh491: Anyways the emscripten Ci would fail for now cause I am playing around with how the shared build for clangCppInteroP would work on emscripten-forge. As I mentioned on discord I wrote my thought on what should be done for cppinterop's emscripten build but I haven't heard anything back on it. Obviously we need to decide on a static build vs a shared build.

vgvassilev: The question is does not having ifdefs around that code cause a failure on emscripten?

anutosh491:
Should be superseded by #199

anutosh491:
#199 is in and hence closing.

PR #191, opened 2024-12-05
Fix github page workflow

mcbarton:
# Description

@anutosh491  The Github page workflow was broken when we stopped using condas emsdk. This PR should fix that.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -37,18 +37,15 @@ jobs:
       - name: Setup default Build Type on *nux
         if: ${{ runner.os != 'windows' }}
         run: |
-          os="${{ matrix.os }}"
-          if [[ "${os}" == "macos"* ]]; then
-            echo "ncpus=$(sysctl -n hw.ncpu)" >> $GITHUB_ENV
-          else
-            echo "ncpus=$(nproc --all)" >> $GITHUB_ENV
-          fi
+          echo "ncpus=$(nproc --all)" >> $GITHUB_ENV
           
-      - name: Setup default Build Type on Windows
-        if: ${{ runner.os == 'windows' }}
+      - name: Setup emsdk
+        shell: bash -l {0}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-github-page.yml
```diff
@@ -37,18 +37,15 @@ jobs:
...
+        shell: bash -l {0}
         run: |
-          $env:ncpus=$([Environment]::ProcessorCount) 
-          echo "ncpus=$env:ncpus" >> $env:GITHUB_ENV
+          cd $HOME
+          git clone https://github.com/emscripten-core/emsdk.git
+          cd emsdk
+          ./emsdk install  ${{ matrix.emsdk_ver }}
 
       - name: Setup emsdk
```
anutosh491, line 50: Can this be removed ? Feels like duplication to me.

anutosh491 (approved):
Mostly looks good except my comment. 
Let's wait for the ci to be green before merging !

.github/workflows/deploy-github-page.yml
mcbarton: This is not duplicating anything as far as I can see. This is a separate workflow and setups emsdk the same way main does.

anutosh491: Wait I am confused. Looking into your file I see this 
```
      - name: Setup emsdk
        shell: bash -l {0}
        run: |
          cd $HOME
          git clone https://github.com/emscripten-core/emsdk.git
          cd emsdk
          ./emsdk install  ${{ matrix.emsdk_ver }}

      - name: Setup emsdk
        shell: bash -l {0}
        run: |
          emsdk install ${{matrix.emsdk_ver}}   
          
```
I am just guessing we don't need to do this twice ?

anutosh491: cc @mcbarton Am I missing something ? Check my comment above

PR #199, opened 2024-12-22
Add Initial Support for xeus-cpp-lite

anutosh491:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -78,8 +78,8 @@ jobs:
           micromamba activate xeus-lite-host
           python -m pip install jupyterlite-xeus
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }} --output-dir dist
-          #cp xcpp.data dist/extensions/@jupyterlite/xeus/static
-          #cp $PREFIX/lib/libclangCppInterOp.so dist/extensions/@jupyterlite/xeus/static
+          cp xcpp.data dist/extensions/@jupyterlite/xeus/static
+          cp $PREFIX/lib/libclangCppInterOp.so dist/extensions/@jupyterlite/xeus/static
           mkdir -p dist/files
           mv notebooks dist/files
           mv README.md dist/files
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -59,6 +59,7 @@ jobs:
           echo "PREFIX=$PREFIX" >> $GITHUB_ENV
           export CMAKE_PREFIX_PATH=$PREFIX
           export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
+          export SYSROOT_PATH=$HOME/emsdk/upstream/emscripten/cache/sysroot
 
           emcmake cmake \
             -DCMAKE_BUILD_TYPE=Release                        \
@@ -66,6 +67,7 @@ jobs:
             -DCMAKE_INSTALL_PREFIX=$PREFIX                    \
             -DXEUS_CPP_EMSCRIPTEN_WASM_BUILD=ON               \
             -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON            \
+            -DSYSROOT_PATH=$SYSROOT_PATH                      \
             ..
           emmake make -j  ${{ env.ncpus }} install
 
@@ -76,6 +78,8 @@ jobs:
           micromamba activate xeus-lite-host
           python -m pip install jupyterlite-xeus
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }} --output-dir dist
+          #cp xcpp.data dist/extensions/@jupyterlite/xeus/static
+          #cp $PREFIX/lib/libclangCppInterOp.so dist/extensions/@jupyterlite/xeus/static
           mkdir -p dist/files
           mv notebooks dist/files
           mv README.md dist/files
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -276,8 +276,8 @@ jobs:
           micromamba activate xeus-lite-host
           python -m pip install jupyterlite-xeus
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }}
-          #cp xcpp.data _output/extensions/@jupyterlite/xeus/static
-          #cp $PREFIX/lib/libclangCppInterOp.so _output/extensions/@jupyterlite/xeus/static
+          cp xcpp.data _output/extensions/@jupyterlite/xeus/static
+          cp $PREFIX/lib/libclangCppInterOp.so _output/extensions/@jupyterlite/xeus/static
 
       - name: Setup tmate session
         if: ${{ failure() && runner.debug }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -257,6 +257,7 @@ jobs:
           echo "PREFIX=$PREFIX" >> $GITHUB_ENV
           export CMAKE_PREFIX_PATH=$PREFIX
           export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
+          export SYSROOT_PATH=$HOME/emsdk/upstream/emscripten/cache/sysroot
 
           emcmake cmake \
             -DCMAKE_BUILD_TYPE=Release                        \
@@ -264,6 +265,7 @@ jobs:
             -DCMAKE_INSTALL_PREFIX=$PREFIX                    \
             -DXEUS_CPP_EMSCRIPTEN_WASM_BUILD=ON               \
             -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON            \
+            -DSYSROOT_PATH=$SYSROOT_PATH                      \
             ..
           emmake make -j  ${{ env.ncpus }} install
 
@@ -274,6 +276,8 @@ jobs:
           micromamba activate xeus-lite-host
           python -m pip install jupyterlite-xeus
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }}
+          #cp xcpp.data _output/extensions/@jupyterlite/xeus/static
+          #cp $PREFIX/lib/libclangCppInterOp.so _output/extensions/@jupyterlite/xeus/static
 
       - name: Setup tmate session
         if: ${{ failure() && runner.debug }}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,6 +520,7 @@ if(EMSCRIPTEN)
     install(FILES
             "$<TARGET_FILE_DIR:xcpp>/xcpp.js"
             "$<TARGET_FILE_DIR:xcpp>/xcpp.wasm"
+            "$<TARGET_FILE_DIR:xcpp>/xcpp.data"
             DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif ()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,10 +152,7 @@ function(configure_kernel kernel)
 endfunction()
 
 message("Configure kernels: ...")
-if(NOT EMSCRIPTEN)
-    configure_kernel("/share/jupyter/kernels/xcpp17/")
-    configure_kernel("/share/jupyter/kernels/xcpp20/")
-else()
+if(EMSCRIPTEN)
     # TODO: Currently jupyterlite-xeus and xeus-lite do not provide
     # methods to fetch information from the arguments present in the
     # generated emscripten kernel.
@@ -166,6 +163,9 @@ else()
     #    to be able to deal with arguments present in kernel.json
     # 3) Finally we should fetch the C++ version from the kernel.json file and
     #    be able to pass it to our wasm interpreter rather than forcing a version.
+    configure_kernel("/share/jupyter/kernels/xcpp20/")
+else()
+    configure_kernel("/share/jupyter/kernels/xcpp17/")
     configure_kernel("/share/jupyter/kernels/xcpp20/")
 endif()
 
@@ -415,6 +415,8 @@ if(EMSCRIPTEN)
     xeus_cpp_set_kernel_options(xcpp)
     xeus_wasm_compile_options(xcpp)
     xeus_wasm_link_options(xcpp "web,worker")
+    # TODO: Remove the exported runtime methods
+    # after the next xeus release.
     target_link_options(xcpp PUBLIC
         -sEXPORTED_RUNTIME_METHODS=FS,PATH,ERRNO_CODES
         # add sysroot location here
@@ -423,7 +425,7 @@ if(EMSCRIPTEN)
     # TODO: Emscripten supports preloading files just once before it generates
     # the xcpp.data file (containing the binary representation of the file(s) we
     # want to include in our application).
-    # Hence although we are adding supporting for Standard Headers, Libraries etc
+    # Hence although we are adding support for Standard Headers, Libraries etc
     # through emscripten's sysroot for now, we need to do the following:
     # 1) Enable CppInterOp to provide us with a resource dir.
     # 2) If the above cannot be done, we can use the resource dir provided
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ if(EMSCRIPTEN)
     set(XEUS_CPP_USE_SHARED_XEUS_CPP OFF)
     set(XEUS_CPP_BUILD_TESTS OFF)
     # ENV (https://github.com/emscripten-core/emscripten/commit/6d9681ad04f60b41ef6345ab06c29bbc9eeb84e0)
-    set(EMSCRIPTEN_FEATURES "${EMSCRIPTEN_FEATURES} -s \"EXTRA_EXPORTED_RUNTIME_METHODS=[ENV']\"")
+    set(EMSCRIPTEN_FEATURES "${EMSCRIPTEN_FEATURES} -s \"EXPORTED_RUNTIME_METHODS=[ENV']\"")
 endif()
 
 # Dependencies
@@ -152,8 +152,22 @@ function(configure_kernel kernel)
 endfunction()
 
 message("Configure kernels: ...")
-configure_kernel("/share/jupyter/kernels/xcpp17/")
-configure_kernel("/share/jupyter/kernels/xcpp20/")
+if(NOT EMSCRIPTEN)
+    configure_kernel("/share/jupyter/kernels/xcpp17/")
+    configure_kernel("/share/jupyter/kernels/xcpp20/")
+else()
+    # TODO: Currently jupyterlite-xeus and xeus-lite do not provide
+    # methods to fetch information from the arguments present in the
+    # generated emscripten kernel.
+    # The following needs to be done here :
+    # 1) We need to configure the kernel properly
+    #    Check issue https://github.com/compiler-research/xeus-cpp/issues/185.
+    # 2) Once the above is done we need to add support in jupyterlite-xeus & xeus-lite
+    #    to be able to deal with arguments present in kernel.json
+    # 3) Finally we should fetch the C++ version from the kernel.json file and
+    #    be able to pass it to our wasm interpreter rather than forcing a version.
+    configure_kernel("/share/jupyter/kernels/xcpp20/")
+endif()
 
 # Source files
 # ============
@@ -401,8 +415,22 @@ if(EMSCRIPTEN)
     xeus_cpp_set_kernel_options(xcpp)
     xeus_wasm_compile_options(xcpp)
     xeus_wasm_link_options(xcpp "web,worker")
+    target_link_options(xcpp PUBLIC
+        -sEXPORTED_RUNTIME_METHODS=FS,PATH,ERRNO_CODES
+        # add sysroot location here
+        --preload-file ${SYSROOT_PATH}/include@/include
+    )
+    # TODO: Emscripten supports preloading files just once before it generates
+    # the xcpp.data file (containing the binary representation of the file(s) we
+    # want to include in our application).
+    # Hence although we are adding supporting for Standard Headers, Libraries etc
+    # through emscripten's sysroot for now, we need to do the following:
+    # 1) Enable CppInterOp to provide us with a resource dir.
+    # 2) If the above cannot be done, we can use the resource dir provided
+    #    by llvm on emscripten-forge but would involve adding a dependency.
+    # 3) Shift the resource dir and the sysroot to a common location.
+    # 4) Preload everything required together.
 endif()
-
 # Tests
 # =====
 
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ pushd build
 export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-host 
 export CMAKE_PREFIX_PATH=$PREFIX
 export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
+export SYSROOT_PATH=$HOME/emsdk/upstream/emscripten/cache/sysroot
 
 emcmake cmake \
         -DCMAKE_BUILD_TYPE=Release                        \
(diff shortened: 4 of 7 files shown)


README.md
```diff
@@ -105,12 +107,20 @@ micromamba activate xeus-lite-host
 python -m pip install jupyterlite-xeus
 jupyter lite build --XeusAddon.prefix=$PREFIX
 ```
+
+We now need to shift necessary files like `xcpp.data` which contains the binary representation of the file(s)
+we want to include in our application. As of now this would contain all important files like Standard Headers,
+Libraries etc coming out of emscripten's sysroot. Assuming we are still inside build we should do the following
+```bash
+cp xcpp.data _output/extensions/@jupyterlite/xeus/static
+cp $PREFIX/lib/libclangCppInterOp.so _output/extensions/@jupyterlite/xeus/static
```
anutosh491, line 116: Though we would always need to copy `xcpp.data` to a location that is accessible by Jupyter Lite (either `_output/xeus/bin` or `_output/extensions/@jupyterlite/xeus/static` that already have access to xcpp.js and xcpp.wasm .......

The second line here shifting `libclangCppInterOp.so` would only be required temporarily. This is only because JupyterLite doesn't deal with shared libs as expected as I think it expects static libs. Hence once we shift to emsdk 3.1.73 and can support a static lib (libclangCppInterOp.a) this would no longer be required !

Hence this is temporary.

CMakeLists.txt
```diff
@@ -401,8 +415,22 @@ if(EMSCRIPTEN)
...
+    # TODO: Emscripten supports preloading files just once before it generates
+    # the xcpp.data file (containing the binary representation of the file(s) we
+    # want to include in our application).
+    # Hence although we are adding supporting for Standard Headers, Libraries etc
+    # through emscripten's sysroot for now, we need to do the following:
+    # 1) Enable CppInterOp to provide us with a resource dir.
+    # 2) If the above cannot be done, we can use the resource dir provided
+    #    by llvm on emscripten-forge but would involve adding a dependency.
+    # 3) Shift the resource dir and the sysroot to a common location.
+    # 4) Preload everything required together.
```
anutosh491, line 432: TODO with respect to preloading raised here. Please let me know of any questions you'll have or this needs to be rephrased.

CMakeLists.txt
```diff
@@ -152,8 +152,22 @@ function(configure_kernel kernel)
...
+    # methods to fetch information from the arguments present in the
+    # generated emscripten kernel.
+    # The following needs to be done here :
+    # 1) We need to configure the kernel properly
+    #    Check issue https://github.com/compiler-research/xeus-cpp/issues/185.
+    # 2) Once the above is done we need to add support in jupyterlite-xeus & xeus-lite
+    #    to be able to deal with arguments present in kernel.json
+    # 3) Finally we should fetch the C++ version from the kernel.json file and
+    #    be able to pass it to our wasm interpreter rather than forcing a version.
+    configure_kernel("/share/jupyter/kernels/xcpp20/")
```
anutosh491, line 169: TODO with respect to supporting more kernels raised here. We would be going ahead with xcpp20 for now but would like to support all.
Please let me know of any questions you'll have or this needs to be rephrased.

CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ if(EMSCRIPTEN)
     set(XEUS_CPP_USE_SHARED_XEUS_CPP OFF)
     set(XEUS_CPP_BUILD_TESTS OFF)
     # ENV (https://github.com/emscripten-core/emscripten/commit/6d9681ad04f60b41ef6345ab06c29bbc9eeb84e0)
-    set(EMSCRIPTEN_FEATURES "${EMSCRIPTEN_FEATURES} -s \"EXTRA_EXPORTED_RUNTIME_METHODS=[ENV']\"")
+    set(EMSCRIPTEN_FEATURES "${EMSCRIPTEN_FEATURES} -s \"EXPORTED_RUNTIME_METHODS=[ENV']\"")
```
anutosh491, line 87: `EXTRA_EXPORTED_RUNTIME_METHODS` had been deprecated in favour of `EXPORTED_RUNTIME_METHODS` since `2.0.18`

.github/workflows/deploy-github-page.yml
```diff
@@ -76,6 +78,8 @@ jobs:
           micromamba activate xeus-lite-host
           python -m pip install jupyterlite-xeus
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }} --output-dir dist
+          #cp xcpp.data dist/extensions/@jupyterlite/xeus/static
+          #cp $PREFIX/lib/libclangCppInterOp.so dist/extensions/@jupyterlite/xeus/static
```
anutosh491, line 82: We need these changes but I am not sure as to how the deploy pages works ?
What is responsible for executing the `jupyter lite serve` command. 

Check the changes made in README.md to understand that these are essential changes and need to go in before serving.

.github/workflows/main.yml
```diff
@@ -274,6 +276,8 @@ jobs:
           micromamba activate xeus-lite-host
           python -m pip install jupyterlite-xeus
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }}
+          #cp xcpp.data _output/extensions/@jupyterlite/xeus/static
+          #cp $PREFIX/lib/libclangCppInterOp.so _output/extensions/@jupyterlite/xeus/static
```
anutosh491, line 280: I don't think we run the `serve` command anywhere in the CI. We need these instructions just before serving. Hence commented out for now as I am not sure.

src/xinterpreter.cpp
```diff
@@ -42,6 +43,7 @@ void* createInterpreter(const Args &ExtraArgs = {}) {
     ClangArgs.push_back("-isystem");
     ClangArgs.push_back(CxxInclude.c_str());
   }
+#endif
```
anutosh491, line 46: These instructions( `DetectResourceDir` and `DetectSystemCompilerIncludePaths `)  can never be put to use as they are based on `exec` which emscripten doesn't plan on supporting now or in future.

anutosh491: **Even if we are okay having errors coming up in the console due to this. They lead to other problems anyways.**

image

src/xinterpreter.cpp
```diff
@@ -357,7 +362,9 @@ __get_cxx_version ()
 
     void interpreter::init_includes()
     {
-        Cpp::AddIncludePath((xeus::prefix_path() + "/include/").c_str());
+#ifndef EMSCRIPTEN
+         Cpp::AddIncludePath((xeus::prefix_path() + "/include/").c_str());
+#endif
```
anutosh491, line 367: This would end up giving up `$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-host/include/` which is of no use to us as it's a local path web assembly would just want to ignore. 

But even if we go for it, it would lead to path lookup errors for now.

image

src/xinterpreter_wasm.cpp
```diff
@@ -17,7 +17,13 @@ namespace xcpp
...
-        : interpreter(0, nullptr)
+        : interpreter(create_args().size(), create_args().data())
     {
     }
+
+    std::vector<const char*> wasm_interpreter::create_args()
+    {
+        static std::vector<const char*> Args = {"-Xclang", "-std=c++20"};
+        return Args;
+    }
```
anutosh491, line 28: Currently we are only presenting an xcpp20 kernel for the users to use by passing the c++ version as 20 to our wasm interpreter. This would be fixed in the future as raised in the TODO.

CMakeLists.txt
```diff
@@ -401,8 +415,22 @@ if(EMSCRIPTEN)
     xeus_cpp_set_kernel_options(xcpp)
     xeus_wasm_compile_options(xcpp)
     xeus_wasm_link_options(xcpp "web,worker")
+    target_link_options(xcpp PUBLIC
+        -sEXPORTED_RUNTIME_METHODS=FS,PATH,ERRNO_CODES
```
anutosh491, line 419: This would also be removed once we update to xeus 6 (a migration is planned soon)

Check https://github.com/jupyter-xeus/xeus/pull/412

CMakeLists.txt
anutosh491: Technically speaking in the long run we would have a better method to load files (and libraries like clad, xsimd etc) dynamically using `mambajs`(https://github.com/emscripten-forge/mambajs) that is currently in the works.

src/xinterpreter.cpp
```diff
@@ -93,12 +96,14 @@ int __get_cxx_version () {
...
       )";
-
-        auto cxx_version = Cpp::Evaluate(code);
-        return std::to_string(cxx_version);
+      auto cxx_version = Cpp::Evaluate(code);
+      return std::to_string(cxx_version);
+#else
+      constexpr int cxx_version = 20;
+      return std::to_string(cxx_version);
+#endif
```
anutosh491, line 104: Repeating again, we are only sticking to one version of the kernel currently for xeus-cpp-lite.

anutosh491:
Also some more context.

We need a way to store the temporarily generated files (`incr_module_xx.o` and `incr_module_xx.wasm`). 

i) Jupyter Lite uses kind-of a fetch-based FS, except it's an in-browser server proxy. But it acts as if files were coming from a normal Jupyter server.
ii) For temporary/local files we probably shouldn't want to involve the server so using MEMFS makes sense here. 
iii) Hence we move these temporary files to the "/tmp/" directory that emscripten provides by default to store temporary files. 

There might be better methods to address this that I shall look into but for now the most straightforward way to do this is using this patch. (https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/llvm/patches/shift_temporary_files_to_tmp_dir.patch)

And the discussion with Sam Clegg (maintainer of emscripten) and Martin Renou (maintainer of Jupyter Lite) here
https://github.com/emscripten-core/emscripten/issues/23228

Hence to get the above working 
1) I built llvm with the patch
2) I then built cppinterop on top this llvm
3) Used it for xeus-cpp-lite.

anutosh491:
The above would give you the following 


https://github.com/user-attachments/assets/c8eeb845-4d08-4099-986a-d9e459d17660

include/xeus-cpp/xinterpreter_wasm.hpp
```diff
@@ -22,6 +22,9 @@ namespace xcpp
         wasm_interpreter();
         virtual ~wasm_interpreter() = default;
 
+    private:
+
+        static std::vector<const char*> create_args();
```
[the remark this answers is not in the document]
anutosh491: Not required. vector comes out of xinterpreter.hpp/cpp !!

CMakeLists.txt
```diff
@@ -152,8 +152,22 @@ function(configure_kernel kernel)
 endfunction()
 
 message("Configure kernels: ...")
-configure_kernel("/share/jupyter/kernels/xcpp17/")
-configure_kernel("/share/jupyter/kernels/xcpp20/")
+if(NOT EMSCRIPTEN)
```
vgvassilev, line 155:
```suggestion
if(EMSCRIPTEN)
```

please reverse the logic to read easier

CMakeLists.txt
vgvassilev: Add a comment in the code saying that.

include/xeus-cpp/xinterpreter_wasm.hpp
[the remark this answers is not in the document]
vgvassilev: Relying on transitive includes is a bad idea. The suggestion is good.

src/xinterpreter.cpp
```diff
@@ -93,12 +96,14 @@ int __get_cxx_version () {
   }
 __get_cxx_version ()
       )";
-
         auto cxx_version = Cpp::Evaluate(code);
         return std::to_string(cxx_version);
+#else
+        constexpr int cxx_version = 20;
+        return std::to_string(cxx_version);
```
vgvassilev, line 103:
```suggestion
        return "20";
```

src/xinterpreter_wasm.cpp
```diff
@@ -17,7 +17,13 @@ namespace xcpp
 {
 
     wasm_interpreter::wasm_interpreter()
-        : interpreter(0, nullptr)
+        : interpreter(create_args().size(), create_args().data())
```
vgvassilev, line 20:
```suggestion
        : interpreter({"-std=c++20"})
```

vgvassilev (commented):
Thank you for working on this.

src/xinterpreter_wasm.cpp
anutosh491: Hmm not sure we can support as this as per how the interpreter class is defined.

https://github.com/compiler-research/xeus-cpp/blob/a7649c0e901f4618a6ce60df2155716959346288/include/xeus-cpp/xinterpreter.hpp#L33-L38

Hence the probably the best way to do this is to using something to create the arguments ?

src/xinterpreter.cpp
vgvassilev: This can go in `createInterpreter` as `CxxSystemIncludes.push_back(xeus::prefix_path() + "/include/")`

src/xinterpreter_wasm.cpp
vgvassilev: In that case you can set this up in `createInterpreter` in case of Emscripten.

anutosh491: Yeah that something I was going for earlier.

src/xinterpreter.cpp
anutosh491: Hmm kinda confused as to why this should be treated as a system include rather than using `AddIncludePath`

Do we want to have something like `-isystem /Users/anutosh491/micromamba/envs/xeus-cpp/include/` while building xeus-cpp for non wasm cases ?

anutosh491: As we won't be interested in the `init_includes` method as a whole for the wasm case, we could only have it for the non-wasm case at the top itself
```
    interpreter::interpreter(int argc, const char* const* argv) :
       .....
    {
        .....
        init_includes();
        init_preamble();
        init_magic();
    }
```
    
We could only enable it for the non wasm case rather than doing this.
    
```
        void interpreter::init_includes()
    {
#ifndef EMSCRIPTEN
        Cpp::AddIncludePath((xeus::prefix_path() + "/include/").c_str());
#endif
    }
```
    
The same applies for `init_magics` as we have the following (though I strongly believe the file one can be supported in the near future.)
```
    void interpreter::init_magic()
    {
#ifndef EMSCRIPTEN
        preamble_manager["magics"].get_cast<xmagics_manager>().register_magic("xassist", xassist());
        preamble_manager["magics"].get_cast<xmagics_manager>().register_magic("file", writefile());
#endif
    }
```

vgvassilev: > Hmm kinda confused as to why this should be treated as a system include rather than using `AddIncludePath`
> 
> Do we want to have something like `-isystem /Users/anutosh491/micromamba/envs/xeus-cpp/include/` while building xeus-cpp for non wasm cases ?

Isn’t that folder containing other conda packages?

anutosh491: Yes it is but using `AddIncludePath` is enough to add it to the search paths I think.  This is what I see locally !
```

 /Users/anutosh491/micromamba/envs/xeus-cpp/include

 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1

 /Library/Developer/CommandLineTools/usr/lib/clang/15.0.0/include

 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include

 /Library/Developer/CommandLineTools/usr/include

 /Users/anutosh491/micromamba/envs/xeus-cpp/bin/../include/c++/v1

 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)

End of search list.
```

vgvassilev: My suggestion is to avoid spilling ifdefs all over the place but have it in one.

anutosh491: Yes, I get it. Let's do one thing. As we don't expect the init_include, init_magic or init_preamble to work for starters. 

Hence a place we could start with rather than addressing everything separately is this as what I've done in my latest commit.

```
#ifndef EMSCRIPTEN
        init_includes();
        init_preamble();
        init_magic();
#endif
```

And one by one whatever we expect to get working for eg 
1) introspection like `?std::vector`
2) the file magic 

We could start separating those out. And the ones that we don't expect to work for eg `xassist` depends on curl and curl is something that won't be working for wasm as of now .... so these would still be kept to be undefined in case of a wasm build.

anutosh491: My doubt is do we get rid of the init_includes method overall and just shift the functionality to createInterpreter ?

vgvassilev: > This can go in `createInterpreter` as `CxxSystemIncludes.push_back(xeus::prefix_path() + "/include/")`

That' what I am asking in this PR.

anutosh491: > That' what I am asking in this PR.

Actually you know what, the `init_includes` doesn't even play a big part as we think it is !!

That's because it's overlapping with the following in kernel.json.in

https://github.com/compiler-research/xeus-cpp/blob/a7649c0e901f4618a6ce60df2155716959346288/share/jupyter/kernels/xcpp20/kernel.json.in#L12-L13

So the location we are interested in is taken care of anyways as I see the following locally coming out of `kernel.json`
```
 /Users/anutosh491/micromamba/envs/xeus-cpp/include
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1
 /Library/Developer/CommandLineTools/usr/lib/clang/15.0.0/include
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include
 /Library/Developer/CommandLineTools/usr/include
 /Users/anutosh491/micromamba/envs/xeus-cpp/bin/../include/c++/v1
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
End of search list.
```

src/xinterpreter.cpp
```diff
@@ -355,11 +360,6 @@ __get_cxx_version ()
         publish_stream("stderr", s);
     }
 
-    void interpreter::init_includes()
-    {
-        Cpp::AddIncludePath((xeus::prefix_path() + "/include/").c_str());
```
vgvassilev, line 360: This needs to go in createInterpreter as I suggested in previous comment.

anutosh491: I think we should be able to do about it but let's do the above if you think we would need that !

vgvassilev: We had that functionality before and I do not see anything that explains why we should drop it. I think we should drop it generally and solve the problem with parameters to the kernel. If we want that pleas update the commit message and we can drop.

anutosh491: Hmm, maybe check this comment (https://github.com/compiler-research/xeus-cpp/pull/199#discussion_r1896397191)

I do not see any difference including it or not including for the non-wasm cases . I see the following locally with or without it.
```
 "/Users/anutosh491/micromamba/envs/xeus-cpp/bin/xcpp" -cc1 -triple arm64-apple-macosx14.0.0 -Wundef-prefix=TARGET_OS_ -Werror=undef-prefix -Wdeprecated-objc-isa-usage -Werror=deprecated-objc-isa-usage -E -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name "<<< inputs >>>" -mrelocation-model pic -pic-level 2 -pic-is-pie -mframe-pointer=non-leaf -ffp-contract=on -fno-rounding-math -funwind-tables=1 -target-sdk-version=14.4 -fcompatibility-qualified-id-block-type-checking -fvisibility-inlines-hidden-static-local-var -fbuiltin-headers-in-system-modules -fdefine-target-os-macros -target-cpu apple-m1 -target-feature +zcm -target-feature +zcz -target-feature +v8.4a -target-feature +aes -target-feature +altnzcv -target-feature +ccdp -target-feature +complxnum -target-feature +crc -target-feature +dotprod -target-feature +fp-armv8 -target-feature +fp16fml -target-feature +fptoint -target-feature +fullfp16 -target-feature +jsconv -target-feature +lse -target-feature +neon -target-feature +pauth -target-feature +perfmon -target-feature +predres -target-feature +ras -target-feature +rcpc -target-feature +rdm -target-feature +sb -target-feature +sha2 -target-feature +sha3 -target-feature +specrestrict -target-feature +ssbs -target-abi darwinpcs -debugger-tuning=lldb -fdebug-compilation-dir=/Users/anutosh491/work/xeus-cpp/build5 -target-linker-version 951.9 -v -fcoverage-compilation-dir=/Users/anutosh491/work/xeus-cpp/build5 -resource-dir /Users/anutosh491/micromamba/envs/xeus-cpp/lib/clang/19 -isystem /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1 -isystem /Library/Developer/CommandLineTools/usr/lib/clang/15.0.0/include -isystem /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -isystem /Library/Developer/CommandLineTools/usr/include -isystem "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)" -include new -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I /Users/anutosh491/micromamba/envs/xeus-cpp/include -internal-isystem /Users/anutosh491/micromamba/envs/xeus-cpp/bin/../include/c++/v1 -internal-isystem /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/local/include -internal-isystem /Users/anutosh491/micromamba/envs/xeus-cpp/lib/clang/19/include -internal-externc-isystem /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -std=c++20 -fdeprecated-macro -ferror-limit 19 -stack-protector 1 -fblocks -fencode-extended-block-signature -fregister-global-dtors-with-atexit -fgnuc-version=4.2.1 -fno-implicit-modules -fskip-odr-check-in-gmf -fcxx-exceptions -fexceptions -fmax-type-align=16 -fcolor-diagnostics -fincremental-extensions -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o - -x c++ "<<< inputs >>>"
clang -cc1 version 19.1.2 based upon LLVM 19.1.2 default target arm64-apple-darwin23.5.0
ignoring nonexistent directory "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)"
ignoring nonexistent directory "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/local/include"
ignoring nonexistent directory "/Users/anutosh491/micromamba/envs/xeus-cpp/lib/clang/19/include"
ignoring nonexistent directory "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/Library/Frameworks"
ignoring duplicate directory "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"
 /Users/anutosh491/micromamba/envs/xeus-cpp/include
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1
 /Library/Developer/CommandLineTools/usr/lib/clang/15.0.0/include
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include
 /Library/Developer/CommandLineTools/usr/include
 /Users/anutosh491/micromamba/envs/xeus-cpp/bin/../include/c++/v1
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
End of search list.
```

Hence I removed it as we anyways don't need it for the wasm case.

vgvassilev: You won't see a difference with that command as the include path is added at a later stage than printing the invocation commands.

Whatever, if somebody wants to re-enable this, we can solve it in a more consistent way by adding a `-I` in the kernel.json.

vgvassilev (approved):
LGTM!

PR #203, opened 2025-01-09
Remove pinning c++ version from get_stdopt

### LINKED ISSUE #202: Remove pinning `cxx_version` in `get_stdopt` for xeus-cpp-lite
Currently we only support c++ 20 based wasm kernel and this flag is passed to our wasm interpreter here
https://github.com/compiler-research/xeus-cpp/blob/dc142225d320b472e88e212cd21a2d720d3f7fc2/src/xinterpreter.cpp#L29-L33

This is the only place where essentially we should be hardcoing the version `20` before we start supporting multiple wasm kernels.

Currently `get_stdopt` is framed like this 
```
    static std::string get_stdopt()
    {
        // We need to find what's the C++ version the interpreter runs with.
#ifndef EMSCRIPTEN
        const char* code = R"(
int __get_cxx_version () {
#if __cplusplus > 202302L
    return 26;
#elif __cplusplus > 202002L
    return 23;
#elif __cplusplus >  201703L
    return 20;
#elif __cplusplus > 201402L
    return 17;
#elif __cplusplus > 201103L || (defined(_WIN32) && _MSC_VER >= 1900)
    return 14;
#elif __cplusplus >= 201103L
   return 11;
#else
  return 0;
#endif
  }
__get_cxx_version ()
      )";
        auto cxx_version = Cpp::Evaluate(code);
        return std::to_string(cxx_version);
#else
        return "20";
#endif
    }
    
```

But we should simply have 
```
    static std::string get_stdopt()
    {
        // We need to find what's the C++ version the interpreter runs with.
        const char* code = R"(
int __get_cxx_version () {
#if __cplusplus > 202302L
    return 26;
#elif __cplusplus > 202002L
    return 23;
#elif __cplusplus >  201703L
    return 20;
#elif __cplusplus > 201402L
    return 17;
#elif __cplusplus > 201103L || (defined(_WIN32) && _MSC_VER >= 1900)
    return 14;
#elif __cplusplus >= 201103L
   return 11;
#else
  return 0;
#endif
  }
__get_cxx_version()
      )";
        auto cxx_version = Cpp::Evaluate(code);
        return std::to_string(cxx_version);
    }
```

That is basically run the `__get_cxx_version` as a first side module and fetch the version through `Evaluate` rather than hardcode it, cause xeus-cpp-lite should be able to run the `code` from `get_stdopt` anyways.

anutosh491:
# Description

Fixes #202 

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -77,7 +77,6 @@ namespace xcpp
     static std::string get_stdopt()
     {
         // We need to find what's the C++ version the interpreter runs with.
-#ifndef EMSCRIPTEN
         const char* code = R"(
 int __get_cxx_version () {
 #if __cplusplus > 202302L
@@ -100,9 +99,6 @@ __get_cxx_version ()
       )";
         auto cxx_version = Cpp::Evaluate(code);
         return std::to_string(cxx_version);
-#else
-        return "20";
-#endif
     }
 
     interpreter::interpreter(int argc, const char* const* argv) :


anutosh491:
After I updated cppinterop's recipe on emscripten-forge (https://github.com/emscripten-forge/recipes/pull/1515)
we can let `Cpp::Evaluate` capture the LastValue. 

So we don't need to hardcode the version rather, we can run 
```
int __get_cxx_version () {
#if __cplusplus > 202302L
    return 26;
#elif __cplusplus > 202002L
    return 23;
#elif __cplusplus >  201703L
    return 20;
#elif __cplusplus > 201402L
    return 17;
#elif __cplusplus > 201103L || (defined(_WIN32) && _MSC_VER >= 1900)
    return 14;
#elif __cplusplus >= 201103L
   return 11;
#else
  return 0;
#endif
  }
__get_cxx_version ()
```
In the browser itself and fetch `20`. (can be seen if add a simple cout statement in get_stdopt after fetching the version)
```
auto cxx_version = Cpp::Evaluate(code);
std::cout << std::to_string(cxx_version) << std::endl;
return std::to_string(cxx_version); 
```

would give the following

vgvassilev (approved):
LGTM!

anutosh491:
Thank you, merging !

PR #204, opened 2025-01-09
Fix faulty CMAKE_FIND_ROOT_PATH_MODE_PACKAGE used for the wasm build

anutosh491:
# Description

I'm not sure which PR introduced `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` but it is being used wrongly.

[Cmake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_PACKAGE.html#variable:CMAKE_FIND_ROOT_PATH_MODE_PACKAGE) say that there are only 3 values this variables accepts 

1) ONLY
2) NEVER
3) BOTH

What we are using is something like `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON` which doesn't make sense (**its wrong but somehow the build just works**)

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -57,8 +57,6 @@ jobs:
 
           export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-host 
           echo "PREFIX=$PREFIX" >> $GITHUB_ENV
-          export CMAKE_PREFIX_PATH=$PREFIX
-          export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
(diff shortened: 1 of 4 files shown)


anutosh491:
cc @mcbarton for reviews

It was a faulty thing on xeus (xeus-cookiecutter) which was responsible for this. I fixed it on xeus yersterday (https://github.com/jupyter-xeus/xeus/pull/416) 

We need to do the same on libraries that might affect xeus-cpp (I don't think other libraries have this mistake but cppinterop does use  `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ON` which needs to be fixed)

anutosh491:
> you also remove where you export this variable just before building?

Absolutely, that's not required for the build.

PR #208, opened 2025-01-10
Add Windows 2025 server to ci

mcbarton:
# Description

This PR adds a Windows sever 2025 build to the ci.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,6 +45,9 @@ jobs:
           - name: Windows22
             os: windows-2022
             micromamba_shell_init: cmd.exe
+          - name: Windows25
+            os: windows-2025
+            micromamba_shell_init: cmd.exe
 
     steps:
       - uses: actions/checkout@v4


mcbarton:
@anutosh491 Its something I'm working on. It fails as you can't write to standard output and standard error at the moment. Stuff which doesn't require printing still works. I'm hoping to have it fixed sometime this year through CppInterOp. Some work is currently being done in llvm so that everything should work.

anutosh491:
> It fails as you can't write to standard output and standard error at the moment. Stuff which doesn't require printing still works. I'm hoping to have it fixed sometime this year through CppInterOp. 

So what is the CI exactly doing ? Are we testing stuff (the python and the c++ tests) ? 
How is this helping us if we can't get xeus-cpp on windows is what I would like to understand ?
I don't see a point populating the CI if its not really beneficial.

mcbarton:
@anutosh491 Its beneficial since we do python and c++ tests, and this new ci tests that the stuff which does work for Windows server 2022 also works for Windiws 2025 while we wait for the fix to the things which don't currently work.

anutosh491 (approved):
Okay. Let's get this in !

PR #211, opened 2025-01-10
Remove omp variable from non openmp kernels

mcbarton:
# Description

These kernels do not support and openmp (and the plan is that the ones that do will be separate kernels). This PR removes references to openmp in the kernel configurations for the kernels currently being built.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature


anutosh491 (approved):
This looks fine considering we don't support openmp right now !

PR #214, opened 2025-01-11
Remove conda forge channel from xeus-cpp wasm host environment

mcbarton:
# Description

When we call the wasm environment we call it under the emscripten-wasm32 platform which doesn't exist for the conda forge channel. As  a result we get the following error in the build
https://github.com/compiler-research/xeus-cpp/actions/runs/12723529061/job/35468733198?pr=213#step:7:51 . Removing the reference to the conda-forge channel fixes this.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -1,7 +1,6 @@
 name: xeus-cpp-wasm-host
 channels:
   - https://repo.mamba.pm/emscripten-forge
-  - https://repo.mamba.pm/conda-forge
 dependencies:
   - nlohmann_json
   - xeus-lite


environment-wasm-host.yml
```diff
@@ -1,11 +1,10 @@
 name: xeus-cpp-wasm-host
 channels:
   - https://repo.mamba.pm/emscripten-forge
-  - https://repo.mamba.pm/conda-forge
```
anutosh491, line 4: Hmmm, looks good to me but I think this ended up here for a reason. Let me confirm on the xeus side and get back.

mcbarton: @anutosh491 Ok. Just a note that CppInterOps environment only references emscripten-forge.

anutosh491:
Ahh I finally think I remember why we need this (and probably why we should keep it)

It is there to support no-arch packages. We currently don't provide any extra functionality as we are in the initial stage of things. But take xeus-r for example (https://github.com/jupyter-xeus/xeus-r/blob/main/environment-wasm-host.yml#L18-L29) provides so many libraries for graphics, image rendering etc.

So in future we might want to add features where packages might be no-arch. Let's just keep this !

PR #216, opened 2025-01-11
Upgrade jupyter kernel test to <0.8

mcbarton:
# Description

Maybe I'm wrong but I don't feel like we need this restriction on the upper limit for jupyter kernel test.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -16,7 +16,7 @@ dependencies:
   - zlib
   # Test dependencies
   - pytest
-  - jupyter_kernel_test>=0.5,<0.6
+  - jupyter_kernel_test>=0.5
   - papermill
   - nbformat
   - nbval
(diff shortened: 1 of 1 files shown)


environment-dev.yml
```diff
@@ -16,9 +16,9 @@ dependencies:
   - zlib
   # Test dependencies
   - pytest
-  - jupyter_kernel_test>=0.5,<0.6
+  - jupyter_kernel_test>=0.5
```
anutosh491, line 19: I think this is good but 

1) I think there is a reason why a pinning was made to >=0.5,<0.6 for jupyter_kernel_test for all xeus kernels. I'll need some time to verify this. 

2) Even if we want to make this change it doesn't make sense to pin a lower version as anyways a higher version would be used. We need to do `- jupyter_kernel_test<0.8`

Wanted to approve and merge but I'll do it soon once I confirm what's going on with point 1.

anutosh491 (approved):
Really good catch !

Approving. Will merge after CI passes.

PR #224, opened 2025-01-16
Update to emsdk 3.1.73

mcbarton:
# Description

This is a PR to upgrade the emsdk version we use to version 3.1.73, once emscripten forge is finished with its transition to this version.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature


.github/workflows/deploy-github-page.yml
```diff
@@ -22,7 +22,6 @@ jobs:
         include:
           - name: Github-page
             os: ubuntu-24.04
-            emsdk_ver: "3.1.45"
```
anutosh491, line 25: Perfect

environment-wasm-host.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - https://repo.mamba.pm/emscripten-forge
```
anutosh491, line 3: Ahhh we need to use the correct channel here. I think all updated packages built with the new version are being uploaded to 
`https://repo.prefix.dev/emscripten-forge-dev`

See here https://prefix.dev/channels/emscripten-forge-dev/packages/cppinterop

anutosh491:
I expect the build to pass, but we should replicate this locally and try what works as of now (atleast the example notebook again from start) and only then we can ensure this is the change to make. Cause the CI would pass anyways but we don't test the example notebook hence though stuff builds it might not run completely I'm guessing untill we try !

.github/workflows/main.yml
```diff
@@ -198,23 +198,22 @@ jobs:
         include:
           - name: ubu24
             os: ubuntu-24.04-arm
```
anutosh491, line 200: Change to ubuntu-24.04 
for now. I see some error involving mamba here. I've raised it to the mamba folks !

PR #228, opened 2025-01-20
Fix retrieve_tagconf_dir and retrieve_tagfile_dir functions

anutosh491:
# Description

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/src/xutils.cpp
+++ b/src/xutils.cpp
@@ -112,7 +112,7 @@ namespace xcpp
         const char separator = '/';
 #endif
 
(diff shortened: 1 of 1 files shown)


anutosh491:
As simple cout statements here can show 

https://github.com/compiler-research/xeus-cpp/blob/53d2f64ba0e43ac47ae30e99ceb10bf502cf488d/src/xinspect.hpp#L125-L126 

image

That the paths we are using are wrong. We just got lucky here as I think as only the last bit of the file path is being used.

mcbarton (approved):

PR #230, opened 2025-01-20
Fix wasm build instructions

mcbarton:
# Description

This PR fixes issues in the wasm build instructions. The first is that we cd into home to install and activate emsk, then try and activate yml file in xeus-cpp repo, but we are no longer in it. The second is that the micromamba create command doesn't work unless you point it towards the channel needed.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,8 +80,9 @@ You are now in a position to build the xeus-cpp kernel. You build it by executin
 micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32
 mkdir build
 cd build
-export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
-export SYSROOT_PATH=$PREFIX/opt/emsdk/upstream/emscripten/cache/sysroot
+export BUILD_TOOLS_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
+export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-host
+export SYSROOT_PATH=$BUILD_TOOLS_PREFIX/opt/emsdk/upstream/emscripten/cache/sysroot
 
 emcmake cmake \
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -98,7 +95,7 @@ emmake make install
```
anutosh491, line 95: I figured out this flaw in my PR too. Now that we use the patched emsdk we need to use its sysroot !

So we can do 
```
export BUILD_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
export SYSROOT_PATH=$BUILD_PREFIX/opt/emsdk/upstream/emscripten/cache/sysroot
```
The sysroot will be present in the opt (optional folder that takes care of such third party things we need)

CONTRIBUTING.md
```diff
@@ -68,23 +68,20 @@ git clone https://github.com/<your-github-username>/xeus-cpp.git
...
 You are now in a position to build the xeus-cpp kernel. You build it by executing the following
 ```bash
 micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32
 mkdir build
-pushd build
-export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-host 
-export SYSROOT_PATH=$HOME/emsdk/upstream/emscripten/cache/sysroot
+cd build
+export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
+export SYSROOT_PATH=$PREFIX/opt/emsdk/upstream/emscripten/cache/sysroot
```
anutosh491, line 84: Arghh wait, so we want to know the location for the libraries involved like xeus, xeus-lite etc and that would be a part of the PREFIX env !

And what we need as a part of the build, that would be in the BUILD_PREFIX env. Now if we change prefix from "host" to "build" that not let the emcmake cmake command run correct ?

mcbarton: Think this is correct now. Please review.

PR #231, opened 2025-01-22
Fix examples not working in the lite demo notebook

anutosh491:
# Description

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -390,7 +390,12 @@ add_library(xeus-cpp-headers INTERFACE)
 set_target_properties(xeus-cpp-headers PROPERTIES PUBLIC_HEADER "${XCPP_HEADERS}")
 
 if (EMSCRIPTEN)
-    install(TARGETS xeus-cpp-headers PUBLIC_HEADER DESTINATION ${SYSROOT_PATH}/include/xcpp)
+    execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory ${SYSROOT_PATH}/include/xcpp)
+
+    message(STATUS "Moving headers to ${SYSROOT_PATH}/include/xcpp")
+    foreach(header ${XCPP_HEADERS})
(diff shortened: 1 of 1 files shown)


anutosh491:
I realized that some examples from the update lite notebook were not working as expected 

image


Whereas we should see the image here 

image

CMakeLists.txt
```diff
@@ -390,7 +390,12 @@ add_library(xeus-cpp-headers INTERFACE)
 set_target_properties(xeus-cpp-headers PROPERTIES PUBLIC_HEADER "${XCPP_HEADERS}")
 
 if (EMSCRIPTEN)
-    install(TARGETS xeus-cpp-headers PUBLIC_HEADER DESTINATION ${SYSROOT_PATH}/include/xcpp)
+    execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory ${SYSROOT_PATH}/include/xcpp)
+
+    message(STATUS "Moving headers to ${SYSROOT_PATH}/include/xcpp")
+    foreach(header ${XCPP_HEADERS})
+        file(COPY ${header} DESTINATION ${SYSROOT_PATH}/include/xcpp)
+    endforeach()
```
anutosh491, line 398: For this to work, we need to preload the required headers (xdisplay.hpp and xmime.hpp to a location where they are accessible)

I expected the `install` command to do the job here. 

I even tried avoiding the complexity of handling `INTERFACE` targets and potential CMake nuances with `PUBLIC_HEADER` and just tried 
```
install(FILES ${XCPP_HEADERS} DESTINATION ${SYSROOT_PATH}/include/xcpp)
```

But even this isn't doing the job. 

Hence here I am 
1) Creating the required directory inside `${SYSROOT_PATH}/include` that we are already preloading
2) Explicitly copying the header files there.

And this works !! 
I am not sure why the `install` isn't comfortable around emscripten !?
But untill someone provides a better solution, this should do the required.

anutosh491:
Merging to see if this fixes the deployment !

PR #232, opened 2025-01-22
Use emsdk from emscripten-forge for deploy build

anutosh491:
# Description

This Pr starts using the recommended patched emsdk from emscripten-forge for the deploy build.

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -2,15 +2,16 @@ name: Build and Deploy
 
 on:
   workflow_dispatch:
+  pull_request:
   push:
     branches:
     - main
-    
+
 permissions:
   contents: read
   pages: write
   id-token: write
-    
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


environment-wasm-build.yml
```diff
@@ -0,0 +1,7 @@
+name: xeus-r-wasm-build
```
mcbarton, line 1: Looks like you copied this from xeus-r . Needs fixing. Also don't think we need cmake.

.github/workflows/deploy-github-page.yml
```diff
@@ -20,38 +21,30 @@ jobs:
       matrix:
         include:
           - name: Github-page
-            os: macos-15
+            os: ubuntu-24.04
```
mcbarton, line 24: Please change to Ubuntu-24.04-arm

.github/workflows/deploy-github-page.yml
```diff
@@ -2,15 +2,16 @@ name: Build and Deploy
 
 on:
   workflow_dispatch:
+  pull_request:
```
mcbarton, line 5: Please remove otherwise you replicate what is already being done when the main workflow runs.

mcbarton (changes requested):

environment-wasm-build.yml
anutosh491: Thank you !

.github/workflows/deploy-github-page.yml
anutosh491: I am confused. So we do want this Ci to run when we make a PR correct ?

.github/workflows/deploy-github-page.yml
```diff
@@ -89,13 +82,16 @@ jobs:
 
   deploy:
     needs: build
+    if: github.ref == 'refs/heads/main'
```
mcbarton, line 85: This is not redundant now we've removed the pull request part at the top of the workflow.

mcbarton (changes requested):

environment-wasm-build.yml
```diff
@@ -0,0 +1,7 @@
+name: xeus-cpp-wasm-build
+channels:
+  - https://repo.mamba.pm/emscripten-forge
+  - conda-forge
+dependencies:
+  - cmake
```
mcbarton, line 6: Can remove cmake.

mcbarton (changes requested):

environment-wasm-build.yml
anutosh491: My thinking here is we anyways need to activate this environment. Hence in our readme, we ask users to do the same ... and rather than us trusting them to already have cmake installed on their device, we can provide it !!!

mcbarton: I'll put this as resolved, as I'm not too fussed about this change.

.github/workflows/deploy-github-page.yml
anutosh491: I think this puts the build job as the part of the CI and would help us test out deployment build for every PR.

mcbarton (approved):
Approved subject to the ci passing.

PR #246, opened 2025-01-29
Update xeus-cpp-lite to use emsdk 3.1.73

anutosh491:
# Description

Completes the update to 3.1.73 !

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -1,9 +1,8 @@
 name: xeus-cpp-wasm-host
 channels:
-  - https://repo.mamba.pm/emscripten-forge
+  - https://repo.prefix.dev/emscripten-forge-dev
   - https://repo.mamba.pm/conda-forge
 dependencies:
-  - emscripten-abi==3.1.45
   - nlohmann_json
   - xeus-lite
   - xeus
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -199,13 +199,9 @@ jobs:
       matrix:
         include:
           - name: ubu24
-            os: ubuntu-24.04-arm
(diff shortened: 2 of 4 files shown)


anutosh491:
This shows https://github.com/compiler-research/xeus-cpp/actions/runs/13029493410/job/36345418308?pr=246#step:6:38

That we use jupyterlite 0.5.0 yet, we need to  use 0.5.1 here ! 
Maybe use it through pip ?

anutosh491:
Closing and Reopening to re-run jobs

environment-wasm-host.yml
```diff
@@ -1,9 +1,9 @@
 name: xeus-cpp-wasm-host
 channels:
-  - https://repo.mamba.pm/emscripten-forge
+  - https://repo.prefix.dev/emscripten-forge-dev
   - https://repo.mamba.pm/conda-forge
 dependencies:
-  - emscripten-abi==3.1.45
+  - emscripten-abi==3.1.73
```
martinRenou, line 6: Do we really need to pin anything here? I guess `3.1.73` is the only version available anyway.

martinRenou (approved):
Just a question. Otherwise looks good!

environment-wasm-host.yml
anutosh491: Hmmm, yes this shouldn't be necessary I think

martinRenou: I mean, we could even remove this line I guess

anutosh491: Cool, let me try that too !

anutosh491: Merging as approved.

PR #248, opened 2025-01-30
[WIP] supporting cppyy

Vipul-Cariappa:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,6 +213,9 @@ set(XEUS_CPP_SRC
     src/xoptions.cpp
     src/xparser.cpp
     src/xutils.cpp
+
+    # magics
+    src/xmagics/pythonexec.cpp
 )
 
 if(NOT EMSCRIPTEN)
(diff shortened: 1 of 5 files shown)


Vipul-Cariappa:
Requesting @mcbarton & @anutosh491 to have a look at this.
Please let me know if the build instructions do not work.

anutosh491:
Hey Vipul thanks for this !

But as far as I know this was not planned/intended just as of now. As in I know xeus-clang-repl was/is working towards going back and forth between cpp and python and I am not sure we want this in xeus-cpp already ! 

I'll cc @JohanMabille here. Johan so [xeus-clang-repl](https://github.com/compiler-research/xeus-clang-repl) used to support magics which allow us to do something like this. Is this something we want to do to ?

image

anutosh491:
Pasting https://github.com/compiler-research/xeus-clang-repl/issues/26 here just to keep track on relevant discussion on the python magic.

SylvainCorlay:
It seems to me that this kind of functionality should probably live in a plugin to xeus-cpp but not in the core of the project.

PR #251, opened 2025-02-04
Fix clang-format workflow

mcbarton:
# Description

The clang format workflow is broken because it tries to get a git-clang-format by wgetting it. Git-clang-format is not used as far as I can tell from the workflow, so should be safe to just remove the 'Download git-clang-format' section of the workflow.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5,7 +5,6 @@
  *                                                                          *
  * The full license is in the file LICENSE, distributed with this software. *
  ****************************************************************************/
-
 #include <cstddef>
 #include <memory>
 #include <string>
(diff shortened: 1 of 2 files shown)


anutosh491:
Looks okay to me but we know the CI would wail for the emscripten build. I'll look into that first and then we can move this in !

src/main.cpp
```diff
@@ -5,7 +5,6 @@
  *                                                                          *
  * The full license is in the file LICENSE, distributed with this software. *
  ****************************************************************************/
-
```
anutosh491, line 8: Not sure why this space was removed.

anutosh491 (approved):
Looks Okay to me but I've not dealt with clang format work on xeus-cpp in the past so I'll just wait for Vassil or someone else to merge.

vgvassilev (approved):
Lgtm.

PR #261, opened 2025-02-27
Enabled file magic support for xeus-cpp-lite

kr-2003:
# Description
Enabled ```%file``` magic for xeus-cpp-lite. 

# Key Changes
1. Included ```src/xmagics/os.cpp``` while compiling for EMSCRIPTEN in ```CMakeLists.txt```.
2. Registered ```file``` in ```preamble_manager["magics"]```.

# Screenrecord

https://github.com/user-attachments/assets/8af07736-8d45-425a-9d52-c02572277f1b

## Type of change
- [x] New feature


anutosh491:
Hey @tharun571, 

This is exactly what we were talking about yesterday !

1) The file magic works for xeus-cpp
2) We were able to do all sort of file read/write/open/close operations through xeus-cpp-lite. So something like this worked and we could manually create a file already 
```
#include <iostream>
#include <fstream>

std::ofstream outfile ("test.txt");

outfile << "Hello World" << std::endl;

outfile.close();
```
image


3) Which gave me enough confidence that, if we just enable this magic while building for emscripten, things would work right off the bat without any changes

So yeah I just asked Abhinav to check if it works and I think it does work.

anutosh491:
@kr-2003 what I observe is that both our example notebooks (for xeus-cpp and xeus-cpp-lite lack any sort of examples for the %%file magic) 

Would you like to update the notebooks and add something for the both of them ?

anutosh491:
That being said, Tharun do we have any sort of example notebooks for the LLM work ? Not sure I remember, if no maybe we should add one. Probably you can help us address this !

kr-2003:
> @kr-2003 what I observe is that both our example notebooks (for xeus-cpp and xeus-cpp-lite lack any sort of examples for the %%file magic)
> 
> Would you like to update the notebooks and add something for the both of them ?

Yeah. I will do that.

tharun571:
@anutosh491 this makes sense. I tested out XAssist in emscripten but it not seem to work due to the curl functions. Maybe I will test out again. The example is present in the documentation but I will raise a notebook example for general xassist usage as well.

anutosh491:
> I tested out XAssist in emscripten but it not seem to work due to the curl functions. Maybe I will test out again.

So yeah unfortunately this is bound to fail cause libcurl has been marked as "won't work" with emscripten (to my knowledge this means it can be done but involves making changes in emscripten that does not aligns with the roadmap the maintainers have). So yeah unfortunately this way might not work. But that being said [JupyterLite/AI](https://github.com/jupyterlite/ai) is up and we might want to think of a way to connect our magic with this !


 > The example is present in the documentation but I will raise a notebook example for general xassist usage as well.
 
 Ahh yess, I forgot about it. If we have it documentation at one place, let's skip populating the notebook for now. @kr-2003 rather what you can do is ..... you can do something similar, just update these docs https://xeus-cpp.readthedocs.io/en/latest/magics.html on magics adding whatever examples and stating the magic works both on xeus-cpp and lite !

vgvassilev:
This PR needs tests.

anutosh491 (approved):
Looks good

vgvassilev (commented):
@anutosh491, convinced me offline to move forward with this pull request given it has no tests this one time. We should open an issue to remind us we to add tests at our earlier convenience. We should do no more exceptions in landing code with no tests from this point on.

PR #265, opened 2025-03-03
Bump CppInterOp to 1.6.0

anutosh491:
# Description

Bump cppinterop to latest released version

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -10,7 +10,7 @@ dependencies:
   - xeus>=5.0.0
   - xeus-zmq>=3.0,<4.0
   - nlohmann_json=3.11.3
-  - CppInterOp>=1.5.0
+  - CppInterOp>=1.6.0
   - pugixml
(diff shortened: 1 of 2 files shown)


environment-dev.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
   - xeus>=5.0.0
   - xeus-zmq>=3.0,<4.0
   - nlohmann_json=3.11.3
-  - CppInterOp>=1.5.0
```
vgvassilev, line 13: We should drop the version and it should take the newest. It can work with 1.5 too.

anutosh491: Correct. Its was being enforced since the beginning like that I think though shouldn't be needed.

anutosh491: Done. Should be ready now

vgvassilev (approved):
LGTM!

PR #271, opened 2025-03-06
Remove lower bound version on cpp-argparse environment-dev.yml

mcbarton:
# Description

Conda will automatically pick the most recent release of cpp-argparse it can. Therefore the lower bound is not necessary in the yml file. We just need to make sure we don't pick a version 4 release when it happens in case we are not compatible with it.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/docs/source/InstallationAndUsage.rst
+++ b/docs/source/InstallationAndUsage.rst
@@ -22,6 +22,7 @@ this environment with mamba by executing the following
 .. code-block:: bash
 
     micromamba create -f environment-dev.yml
+    micromamba activate xeus-cpp
     micromamba install jupyterlab -c conda-forge
 
 Now you can compile the kernel from the source by executing (replace `$CONDA_PREFIX` 


anutosh491:
Hmm, I guess we don't need anything. We don't have any bound on our feedstock

https://github.com/conda-forge/xeus-cpp-feedstock/blob/main/recipe/meta.yaml#L30

If we think any update would bring any failure (like any deprecation or anything) 
in that case we can probably make a release with the last compatible argparse and then shift to the later version.

anutosh491:
Let's just remove any bound on argparse from everywhere 

https://github.com/search?q=repo%3Acompiler-research%2Fxeus-cpp%20%3E%3D3.0%2C%3C4.0&type=code

anutosh491:
Perfect, let's move it in !

README.md
```diff
@@ -130,9 +130,14 @@ http://xeus-cpp.readthedocs.io
...
 
+| `xeus-cpp` | `xeus-zmq`      | `CppInterOp` | `pugixml` | `nlohmann_json` |
+|------------|-----------------|--------------|-----------|-----------------|
+|  main      |  >=3.0.0,<4.0.0 | >=1.5.0      | ~1.8.1    | >=3.11.3,<4.0   |
+
+Versions `0.6.0` and prior had a restriction on the exact version of cpp-argparse used
+
 | `xeus-cpp` | `xeus-zmq`      | `CppInterOp` | `pugixml` | `cpp-argparse`| `nlohmann_json` |
 |------------|-----------------|--------------|-----------|---------------|-----------------|
-|  main      |  >=3.0.0,<4.0.0 | >=1.5.0      | ~1.8.1    | >=3.0,<4.0    | >=3.11.3,<4.0   |
```
anutosh491, line 135: Wait this table isn't just meant to show restrictions correct ? It meant to show whatever dependencies we currently use too ! So creating a new table and removing cpp-argparse from there doesn't make sense ? We just need to paste whatever latest cpp-argparse is being used (which can be dynamic if there is any update) so I think >=3,<4 does the job isn't it ?

vgvassilev: I always considered that table as a weird way of expressing a list of dependencies and their constrains. Either we drop the entire table or we keep it and keep arg-parse in there. We are not entirely correct that arg-parse is version agnostic because some time ago in some version they introduced a breaking change...

mcbarton: As one of the suggestions I've removed the table in its entirety. I very much doubt anyone is going to go back and build one of these older releases, and if they are, they could just use the environment file in the repo as a guideline.

vgvassilev: I have no preferences.

vgvassilev (approved):
LGTM!

README.md
```diff
@@ -187,18 +188,7 @@ http://xeus-cpp.readthedocs.io
```
anutosh491, line 188: Add pugixml here.

README.md
```diff
@@ -44,7 +44,8 @@ source activate  "xeus-cpp"
 ```
 We will now install the dependencies needed to compile xeux-cpp from source within this environment by executing the following
 ```bash
-mamba install notebook cmake cxx-compiler xeus-zmq nlohmann_json=3.11.3 jupyterlab CppInterOp cpp-argparse">=3.0,<4.0" pugixml doctest -c conda-forge
+micromamba create -f environment-dev.yml
+micromamba install jupyterlab -c conda-forge
```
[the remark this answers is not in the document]
mcbarton: Done I think

PR #272, opened 2025-03-06
Remove lower bound on xeus-zmq version environment-dev.yml

mcbarton:
# Description

Conda will automatically pick the most recent release of xeus-zmq it can. Therefore the lower bound is not necessary in the yml file. We just need to make sure we don't pick a version 4 release when it happens in case we are not compatible with it.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -8,7 +8,7 @@ dependencies:
   - cxx-compiler=1.7.0
   # Host dependencies
   - xeus>=5.0.0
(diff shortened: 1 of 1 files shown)


anutosh491:
Cool, you're not wrong (I just like keeping stuff consistent with the feedstock which maintains >3<4) let's take is in.

anutosh491 (approved):

PR #273, opened 2025-03-06
Remove XEUS_CPP_INCLUDE_DOCS in CMakeLists.txt

mcbarton:
# Description

As the cmake is not currently designed to build the docs despite having XEUS_CPP_INCLUDE_DOCS defined this PR removes the section which mentions this variable. The variable is assumed to be part of an old PR which never got spotted before merge.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -551,7 +551,3 @@ if(EMSCRIPTEN)
             "$<TARGET_FILE_DIR:xcpp>/xcpp.data"
             DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif ()
-
-if(XEUS_CPP_INCLUDE_DOCS)
-    add_subdirectory(docs)
-endif()


anutosh491:
Okay wait, this gets me very confused though :\

If you see the cmake, we have 

https://github.com/compiler-research/xeus-cpp/blob/ec63cc3329ce79bf337da6236ba519dc9ee9b09c/CMakeLists.txt#L556-L558

But this would want a CMakeLists.txt inside the docs folder correct ? (And we don't have that)

```
(xeus-cpp) anutosh491@arm64-apple-darwin20 xeus-cpp % mkdir build
cd build
cmake .. -D CMAKE_PREFIX_PATH=$CONDA_PREFIX -D CMAKE_INSTALL_PREFIX=$CONDA_PREFIX -D CMAKE_INSTALL_LIBDIR=lib -D XEUS_CPP_INCLUDE_DOCS=ON
make install
-- The C compiler identification is Clang 16.0.6
-- The CXX compiler identification is Clang 16.0.6
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Users/anutosh491/micromamba/envs/xeus-cpp/bin/arm64-apple-darwin20.0.0-clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Users/anutosh491/micromamba/envs/xeus-cpp/bin/arm64-apple-darwin20.0.0-clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building xeus-cpp v0.7.0dev
-- Found nlohmann_json: /Users/anutosh491/micromamba/envs/xeus-cpp/share/cmake/nlohmann_json/nlohmann_jsonConfig.cmake (found suitable exact version "3.11.3")
-- Found compatible xeus version: 5.2.0
-- Found CppInterOp: config=/Users/anutosh491/micromamba/envs/xeus-cpp/lib/cmake/CppInterOp dir=/Users/anutosh491/micromamba/envs/xeus-cpp (found version=1.6.0 compatible with Clang 19.x)
Configure kernels: ...
-- Performing Test HAS_CPP_17_FLAG
-- Performing Test HAS_CPP_17_FLAG - Success
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found OpenSSL: /Users/anutosh491/micromamba/envs/xeus-cpp/lib/libcrypto.dylib (found version "3.4.1")
-- Performing Test HAS_MARCH_NATIVE
-- Performing Test HAS_MARCH_NATIVE - Success
CMake Error at CMakeLists.txt:556 (add_subdirectory):
  The source directory

    /Users/anutosh491/work/xeus-cpp/docs

  does not contain a CMakeLists.txt file.


-- Configuring incomplete, errors occurred!
make: *** No rule to make target 'install'.  Stop.
```

So I don't even think we are building our docs dependening on this config here !!

Let's get rid of the following here.
```
if(XEUS_CPP_INCLUDE_DOCS)
    add_subdirectory(docs)
endif()
```

anutosh491:
Hmmm, let's hear what @JohanMabille has to say here. 

I haven't seen docs being built this way on the other xeus kernels I've worked with so a bit confused here. But yeah probably would demand a `CMakeLists.txt` in the docs if we use this.

JohanMabille:
Hi, I don't know why `XEUS_CPP_INCLUDE_DOCS` was initially added to the CMakeLists.txt, maybe it was part of a bigger refactoring / change that has not been fully merged, or a leftover? @mvassilev presumably knows more about it.

The current mechanism for building the docs is independent from cmake, one needs to run `make html` / `make pdf` / ... under the docs folder after installing the [doc environment](https://github.com/jupyter-xeus/xeus-cling/blob/main/docs/environment.yml). The way it is done is "historical" (i.e. I don't remember the details, although I think it is more or less related to readthedocs constraints ;)) and is consistent with the builds of the docs of all the other xeus repos. Not that I am against changing it, but as @mcbarton pointed it out in his last comment, it would probably require more changes to the CMakeLists.txt to get it work. Another option could be to keep the current docs build mechanism and totally remove the `XEUS_CPP_INCLUDE_DOCS` from the CMakeLists to avoid confusion in te future.

anutosh491:
Thanks for the reply.

> Another option could be to keep the current docs build mechanism and totally remove the XEUS_CPP_INCLUDE_DOCS from the CMakeLists to avoid confusion in the future.

Yeah that's what was looking reasonable to me and hence I wrote this in my [comment](https://github.com/compiler-research/xeus-cpp/pull/273#issuecomment-2705656290) above

> So I don't even think we are building our docs dependening on this config here !!

>Let's just get rid of the following here.

> if(XEUS_CPP_INCLUDE_DOCS)
>    add_subdirectory(docs)
> endif()


Maybe let's do this for now ? 
And even if we want to take the cmake route, we can do it through a subsequent pr but atleast lets not have any wrong-doing/leftover from previous work for starters ?

mcbarton:
> Thanks for the reply.
> 
> > Another option could be to keep the current docs build mechanism and totally remove the XEUS_CPP_INCLUDE_DOCS from the CMakeLists to avoid confusion in the future.
> 
> Yeah that's what was looking reasonable to me and hence I wrote this in my [comment](https://github.com/compiler-research/xeus-cpp/pull/273#issuecomment-2705656290) above
> 
> > So I don't even think we are building our docs dependening on this config here !!
> 
> > Let's just get rid of the following here.
> 
> > if(XEUS_CPP_INCLUDE_DOCS)
> > add_subdirectory(docs)
> > endif()
> 
> Maybe let's do this for now ? And even if we want to take the cmake route, we can do it through a subsequent pr but atleast lets not have any wrong-doing/leftover from previous work for starters ?

@anutosh491 Ive remove the XEUS_CPP_INCLUDE_DOCS section. Can you approve now?

anutosh491 (approved):
Perfect

PR #274, opened 2025-03-06
Install graphviz docs environment.yml

mcbarton:
# Description

If you visit the xeus-cpp read the docs build log here https://readthedocs.org/projects/xeus-cpp/builds/27398561/ you'll see many errors about dot not being found and not being able to do certain things. This is fixed if you install Graphviz like it mentions later on in the logs.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -5,5 +5,6 @@ channels:
 
 dependencies:
   - breathe
+  - graphviz
   - pip:
     - sphinx_rtd_theme


mcbarton:
@anutosh491 @JohanMabille @vgvassilev pinging for review.

vgvassilev (approved):
LGTM, however, I am wondering if there is some way to say this is a dependency only for the documentation -- I do not see how but would have been good to be able to express that...

mcbarton:
> LGTM, however, I am wondering if there is some way to say this is a dependency only for the documentation -- I do not see how but would have been good to be able to express that...

This is only added as a dependency of the documentation already. You cannot build the documentation as part of the cmake build process. If you look at the file I edited you'll see that it is a environment file just for the docs folder.

mcbarton:
> LGTM, however, I am wondering if there is some way to say this is a dependency only for the documentation -- I do not see how but would have been good to be able to express that...

Merging, as been approved, and the question about adding as a documentation dependency is what we are already doing.

PR #277, opened 2025-03-07
Enable C++ tests for the Emscripten build

anutosh491:
# Description

Enables C++ tests for the Emscripten build

## Type of change

- [ ] Bug fix
- [x] New feature


CMakeLists.txt
```diff
@@ -82,7 +82,7 @@ if(EMSCRIPTEN)
     set(XEUS_CPP_BUILD_EXECUTABLE OFF)
     set(XEUS_CPP_USE_SHARED_XEUS OFF)
     set(XEUS_CPP_USE_SHARED_XEUS_CPP OFF)
-    set(XEUS_CPP_BUILD_TESTS OFF)
+    set(XEUS_CPP_BUILD_TESTS ON)
```
anutosh491, line 85: ON by default probably no need to set it again. 
Can keep it maybe to keep track !

test/CMakeLists.txt
```diff
@@ -48,22 +50,54 @@ set(XEUS_CPP_TESTS
...
+
+    target_compile_options(test_xeus_cpp
+        PUBLIC "SHELL: -fexceptions"
+    )
+
+    target_link_options(test_xeus_cpp
+        PUBLIC "SHELL: -fexceptions"
+        PUBLIC "SHELL: -s MAIN_MODULE=1"
+        PUBLIC "SHELL: -s ASSERTIONS=0"
+        PUBLIC "SHELL: -s ALLOW_MEMORY_GROWTH=1"
```
anutosh491, line 64: Needed otherwise different memory allocated errors 

image

test/CMakeLists.txt
```diff
@@ -48,22 +50,54 @@ set(XEUS_CPP_TESTS
...
+        PUBLIC "SHELL: -s EXIT_RUNTIME=1"
```
anutosh491, line 65: Needed for doctests to exit smoothly. Otherwise hangs after finishing running the tests.

test/CMakeLists.txt
```diff
@@ -48,22 +50,54 @@ set(XEUS_CPP_TESTS
...
+        PUBLIC "SHELL: --preload-file ../${XEUS_CPP_CONF_DIR}@/etc/xeus-cpp"
+    )
+
+    target_include_directories(test_xeus_cpp PRIVATE ${XEUS_CPP_INCLUDE_DIR})
+
+    add_custom_command(TARGET test_xeus_cpp POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy
+        ${CMAKE_INSTALL_PREFIX}/lib/libclangCppInterOp.so
+        ${CMAKE_CURRENT_BINARY_DIR}/libclangCppInterOp.so
+        COMMENT "Copying libclangCppInterOp.so to the test directory"
```
anutosh491, line 79: Needed because we build cppinterop as a side_module right now and it should be present in the test directory so that it can be loaded on top of the main module (test_xeus_cpp.wasm) . So copying it from prefix to test dir right now.

So technically all 3 files 

test_xeus_cpp.wasm, test_xeus_cpp.data and libclangCppInterOp.so should be present in the same dir.

test/test_interpreter.cpp
```diff
@@ -980,6 +993,7 @@ TEST_SUITE("xassist"){
     }
 
 }
+#endif
```
anutosh491, line 996: Skipping this test as a whole because xassist won't work for lite untill curl is supported by emscripten (which I think has been marked as "won't-fix" by maiantainers. So this isn't expected to work !

test/test_interpreter.cpp
```diff
@@ -613,7 +619,13 @@ TEST_SUITE("xsystem_clone")
 
 TEST_SUITE("xsystem_apply")
 {
+#if defined(XEUS_CPP_EMSCRIPTEN_WASM_BUILD)
+    TEST_CASE("apply_xsystem"
+            * doctest::should_fail(true)
+            * doctest::description("TODO: Currently fails for the Emscripten build"))
+#else
```
anutosh491, line 626: Using this framework (doctest decorators) because we can add details like this fails right now but is expected to work and can be lablled as a TODO !

Gives us this 
```
===============================================================================
/home/runner/work/xeus-cpp/xeus-cpp/test/test_interpreter.cpp:625:
DESCRIPTION: TODO: Currently fails for the Emscripten build
TEST SUITE: xsystem_apply
TEST CASE:  apply_xsystem

/home/runner/work/xeus-cpp/xeus-cpp/test/test_interpreter.cpp:636: FATAL ERROR: REQUIRE( kernel_res["status"] == "ok" ) is NOT correct!
  values: REQUIRE( "error" == ok )

Failed as expected so marking it as not failed
```
And finally we would have 
```
===============================================================================
[doctest] test cases: 45 | 45 passed | 0 failed | 0 skipped
[doctest] assertions: 68 | 66 passed | 2 failed |
[doctest] Status: SUCCESS!
```

anutosh491: So 2 tests fail but overall its a suceess cause we explictily mark these tests as "should-fail"

anutosh491:
Currently we have around 45/47 test working for lite as compared to the native cases. According my surface level debug, these don't look like lite error, rather some internal functions going off I suppose :\ (So not something really major)

But yeah that being said this PR is just to get us started and I have marked the two tests as TODO and shall address them through subsequent prs.

I have tried to be as verbose as possible with the comments above and probably should make it easier to review.

cc @DerThorsten @JohanMabille could you possibly help me with the reviews here ?

test/test_interpreter.cpp
```diff
@@ -169,7 +169,13 @@ TEST_SUITE("execute_request")
 
 TEST_SUITE("inspect_request")
 {
+#if defined(XEUS_CPP_EMSCRIPTEN_WASM_BUILD)
+    TEST_CASE("good_status"
+            * doctest::should_fail(true)
+            * doctest::description("TODO: Currently fails for the Emscripten build"))
+#else
     TEST_CASE("good_status")
+#endif
```
anutosh491, line 178: This fails currently due to something. But its some internal function that might be going off i guess, cause we can run an inspect request and we test it too !!!

https://github.com/compiler-research/xeus-cpp/pull/277/files#diff-174aaa9b249ba194edc94c41ccce44e35f2309dfc519ae8fc44606a6fedec476R101

vgvassilev:
We should merge this PR only after https://github.com/compiler-research/CppInterOp/pull/483 is in.

test/test_interpreter.cpp
DerThorsten: I think the system test can be disabled for for emscripten since we just canno make these system calls.
I would suggest to have the #ifnot emscripten block arround the whole test

DerThorsten (approved):

vgvassilev:
> > We should merge this PR only after [compiler-research/CppInterOp#483](https://github.com/compiler-research/CppInterOp/pull/483) is in.
> 
> The two PRs seem fairly independent. We should move forward with this one IMO rather than creating artificial blockers.

Let’s discuss that on Monday.

test/CMakeLists.txt
```diff
@@ -25,7 +25,10 @@ include(CheckCXXCompilerFlag)
 string(TOUPPER "${CMAKE_BUILD_TYPE}" U_CMAKE_BUILD_TYPE)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES Clang OR CMAKE_CXX_COMPILER_ID MATCHES GNU OR CMAKE_CXX_COMPILER_ID MATCHES Intel)
-    add_compile_options(-Wunused-parameter -Wextra -Wreorder -Wconversion -Wsign-conversion)
+
+    if(NOT EMSCRIPTEN)
```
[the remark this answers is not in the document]
anutosh491: Cause they are no required while building the main module

We never used any of these flags while building the other main module too (xcpp.wasm)

check 

https://github.com/compiler-research/xeus-cpp/blob/48767d6274334f8223532b0676beea492dc0fe6d/CMakeLists.txt#L257-L261

.github/workflows/main.yml
```diff
@@ -247,11 +247,18 @@ jobs:
             ..
           emmake make -j ${{ env.ncpus }} install
 
-      - name: Jupyter Lite integration 
+      - name: Test xeus-cpp C++ Emscripten
         shell: bash -l {0}
         run: |
           micromamba create -n xeus-lite-host jupyterlite-core
```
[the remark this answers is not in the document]
anutosh491: Hmmm, let's just keep it this way ?

Cause this environment provides us with node and anyways we need to stay in this env for the jupyter lite integration !

test/CMakeLists.txt
```diff
@@ -48,22 +50,54 @@ set(XEUS_CPP_TESTS
...
-        MACOSX_RPATH ON
+if(EMSCRIPTEN)
+    target_link_libraries(test_xeus_cpp PRIVATE xeus-cpp-static doctest::doctest)
+
+    target_compile_options(test_xeus_cpp
+        PUBLIC "SHELL: -fexceptions"
+    )
+
+    target_link_options(test_xeus_cpp
+        PUBLIC "SHELL: -fexceptions"
```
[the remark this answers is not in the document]
anutosh491: Arghh no those are absolutely different issues 😅 

A try catch or throw not working at runtime while running clang-repl in the browser is because our clang-repl cc1 command is missing these backend flags 
```
-cc1 -triple wasm32-unknown-emscripten ........ -mllvm -enable-emscripten-cxx-exceptions -mllvm -enable-emscripten-sjljfincremental-extensions -o "<<< inputs >>>.o" -x c++ "<<< inputs >>>"
```

These would be handled through a PR on LLVM (shall paste it on discord when I raise one)

These -fexceptions used at build time and link time play the exact same role as the build time and link time -fexceptions flag used for builing xcpp.wasm (we build the whole stack xeus to xeus-cpp to cppinterop to llvm with -fexceptions but this doesn't overlap with the runtime exception handling error we've caught)

test/CMakeLists.txt
```diff
@@ -48,22 +50,54 @@ set(XEUS_CPP_TESTS
...
+        PUBLIC "SHELL: -s MAIN_MODULE=1"
+        PUBLIC "SHELL: -s ASSERTIONS=0"
+        PUBLIC "SHELL: -s ALLOW_MEMORY_GROWTH=1"
+        PUBLIC "SHELL: -s EXIT_RUNTIME=1"
+        PUBLIC "SHELL: -s STACK_SIZE=32mb"
+        PUBLIC "SHELL: -s INITIAL_MEMORY=128mb"
```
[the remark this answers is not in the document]
anutosh491: No special reason, only to add consistency on top our main module as per how we build xcpp.wasm

Technically we want to run all these things through the xeus-cpp-lite and it only makes sense to use the exact same flags as what we use while building xcpp.wasm

test/CMakeLists.txt
```diff
@@ -48,22 +50,54 @@ set(XEUS_CPP_TESTS
...
+        PUBLIC "SHELL: -fexceptions"
+    )
+
+    target_link_options(test_xeus_cpp
+        PUBLIC "SHELL: -fexceptions"
+        PUBLIC "SHELL: -s MAIN_MODULE=1"
+        PUBLIC "SHELL: -s ASSERTIONS=0"
+        PUBLIC "SHELL: -s ALLOW_MEMORY_GROWTH=1"
+        PUBLIC "SHELL: -s EXIT_RUNTIME=1"
+        PUBLIC "SHELL: -s STACK_SIZE=32mb"
```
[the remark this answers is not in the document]
anutosh491: https://github.com/compiler-research/xeus-cpp/pull/277#discussion_r1986631098

anutosh491:
@mcbarton I think you've had some doubts with flags (like -fexceptions, memory growth etc etc)

All are introduced to keep consistency with xcpp.wasm (which is the main module finally running and responsible for any output ... so obviously our test's main module should be following the same output)

Here you can find all the initital set of buildtime and link time flags that are responsible for building xcpp.wasm

https://github.com/jupyter-xeus/xeus/blob/main/cmake/WasmBuildOptions.cmake

.github/workflows/main.yml
```diff
@@ -247,11 +247,18 @@ jobs:
...
           micromamba activate xeus-lite-host
+          cd build/test
```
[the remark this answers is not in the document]
anutosh491: Addressed.

anutosh491:
> Unless specifically needed to enable and run the tests I don't think they should be part of this PR. I feel that if they are purely there for consistency purposes then they should be added in a separate PR. Are you able to run a commit which shows tests fail if they aren't included?


Why would that be ? We build xcpp.wasm with the same flags and we technically want to exactly replicate what we see with xeus-cpp-lite. Let's just keep them here. I don't see a point splitting them. I think they are essential.

anutosh491:
> Don't we want the build and deploy workflow also to run these tests, so we can check the nightly build is working?

Hmm, I don't know. This initial PR is to get us started. I think we're good with having tests for it in the CI right now. Let's extend it further if we think it might really essential.

anutosh491:
> I don't see a point splitting them. I think they are essential.

okay wow I think when I say "essential", these are really essential. For example getting rid of the stack size/initial memory or changing them not to match xcpp.wasm gives me this 
```
RuntimeError: memory access out of bounds
    at wasm://wasm/00a92bfe:wasm-function[3086]:0x1bab7a
    at wasm://wasm/00a92bfe:wasm-function[3913]:0x1d9bce
    at libclangCppInterOp.so.std::__2::__tree<std::__2::__value_type<clang::QualType, clang::CXXBaseSpecifier*>, std::__2::__map_value_compare<clang::QualType, std::__2::__value_type<clang::QualType, clang::CXXBaseSpecifier*>, clang::QualTypeOrdering, true>, std::__2::allocator<std::__2::__value_type<clang::QualType, clang::CXXBaseSpecifier*>>>::destroy(std::__2::__tree_node<std::__2::__value_type<clang::QualType, clang::CXXBaseSpecifier*>, void*>*) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[15584]:0x115f733)
    at libclangCppInterOp.so.clang::Sema::AttachBaseSpecifiers(clang::CXXRecordDecl*, llvm::MutableArrayRef<clang::CXXBaseSpecifier*>) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[15582]:0x115f59c)
    at libclangCppInterOp.so.clang::Sema::InstantiateClass(clang::SourceLocation, clang::CXXRecordDecl*, clang::CXXRecordDecl*, clang::MultiLevelTemplateArgumentList const&, clang::TemplateSpecializationKind, bool) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[19427]:0x15d7cd7)
    at libclangCppInterOp.so.clang::Sema::InstantiateClassTemplateSpecialization(clang::SourceLocation, clang::ClassTemplateSpecializationDecl*, clang::TemplateSpecializationKind, bool) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[19433]:0x15d9707)
    at libclangCppInterOp.so.void llvm::function_ref<void ()>::callback_fn<clang::Sema::RequireCompleteTypeImpl(clang::SourceLocation, clang::QualType, clang::Sema::CompleteTypeKind, clang::Sema::TypeDiagnoser*)::$_0>(long) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[19850]:0x166071e)
    at libclangCppInterOp.so.clang::Sema::runWithSufficientStackSpace(clang::SourceLocation, llvm::function_ref<void ()>) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[13559]:0xf9512f)
    at libclangCppInterOp.so.clang::Sema::RequireCompleteTypeImpl(clang::SourceLocation, clang::QualType, clang::Sema::CompleteTypeKind, clang::Sema::TypeDiagnoser*) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[19781]:0x16494d4)
    at libclangCppInterOp.so.clang::Sema::RequireCompleteType(clang::SourceLocation, clang::QualType, clang::Sema::CompleteTypeKind, clang::Sema::TypeDiagnoser&) (wasm://wasm/libclangCppInterOp.so-10edf8ce:wasm-function[19765]:0x16465d6)

Node.js v23.9.0
```

.github/workflows/main.yml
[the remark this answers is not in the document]
anutosh491: > each stage has to have it manually activated is my understanding

Hmm is it ?
For eg even for the build we activate xeus-cpp-wasm-build through 
```
      - name: install mamba
        uses: mamba-org/setup-micromamba@main
        with:
          environment-file: environment-wasm-build.yml
          init-shell: bash
          environment-name: xeus-cpp-wasm-build
          
```
and  we do the build next within this env correct ?

> Also we get node from the emsdk installation.

This gives us Node 16, I want to use the latest version possible Node 23 which is present in xeus-lite-host.

.github/workflows/main.yml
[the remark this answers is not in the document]
anutosh491: This was reversed cause I need to "cd" into the test dir because test_xeus_cpp.data and our side module libclangCppInterOp.so lies there. Which needs to be accessed by test_xeus_cpp.wasm !

.github/workflows/main.yml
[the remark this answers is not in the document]
anutosh491: Also I am sure we are in xeus-lite-host and that doesn't need to be activated separately once done. 

Otherwise the following Jupyterlite integration won't even work isn't it ?

https://github.com/compiler-research/xeus-cpp/actions/runs/13759790519/job/38473305453?pr=277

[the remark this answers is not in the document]
anutosh491: Arghhh yes, I think i am wrong here and what you say is correct !

I shall make this change. Thanks a lot for pointing this out !

.github/workflows/main.yml
```diff
@@ -247,10 +247,18 @@ jobs:
             ..
           emmake make -j ${{ env.ncpus }} install
 
-      - name: Jupyter Lite integration 
+      - name: Test xeus-cpp C++ Emscripten
         shell: bash -l {0}
         run: |
           micromamba create -n xeus-lite-host jupyterlite-core
```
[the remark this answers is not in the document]
anutosh491: Hmm. Okay let me try this out.

test/CMakeLists.txt
```diff
@@ -48,22 +50,54 @@ set(XEUS_CPP_TESTS
...
+if(EMSCRIPTEN)
+    target_link_libraries(test_xeus_cpp PRIVATE xeus-cpp-static doctest::doctest)
+
+    target_compile_options(test_xeus_cpp
+        PUBLIC "SHELL: -fexceptions"
+    )
+
+    target_link_options(test_xeus_cpp
+        PUBLIC "SHELL: -fexceptions"
+        PUBLIC "SHELL: -s MAIN_MODULE=1"
```
mcbarton, line 62: @anutosh491 You need

``` 
PUBLIC "SHELL: -s WASM_BIGINT"
```

also. If you don't include this then when someone builds CppInterOp outside of Emscripten forge then you get the following error when trying to run the tests


node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^

[LinkError: WebAssembly.instantiate(): Import #19 module="env" function="_ZNSt3__213basic_ostreamIcNS_11char_traitsIcEEElsEy": imported function does not match the expected type]

Node.js v22.14.0

test/test_interpreter.cpp
mcbarton: This should be skipped rather than ignored completely I believe. This issue for curl was labelled as won't fix, but by bot https://github.com/emscripten-core/emscripten/issues/4906 .  This issue https://github.com/emscripten-core/emscripten/issues/13025 that got a reply didn't get a response that curl doesn't work, just that sockets is very limited. Even going away from curl xeus-cpp could make use of Emscriptens fetch API https://emscripten.org/docs/api_reference/fetch.html .

CMakeLists.txt
mcbarton: This options is set to on by default. It doesn't need to be set here anymore.

mcbarton (changes requested):

mcbarton:
@anutosh491 Please update the documentation as part of this PR so that developers know that xeus-cpp has Emscripten tests, and how to run them.

test/CMakeLists.txt
anutosh491: That's a default option when you build with emsdk from emscripten-forge. Shouldn't be needed here ;)
If you build with verbose logging you shall see it is an option during the linking.

Resolving !

anutosh491:
> https://github.com/compiler-research/xeus-cpp/pull/277#discussion_r1999485500

Let's not stress too much on this. 

1) I just know we can't use curl for sure ( I tried build

anutosh491:
> This should be skipped rather than ignored completely I believe. This one issue for curl was labelled as won't fix, but by bot https://github.com/emscripten-core/emscripten/issues/4906 . This issue https://github.com/emscripten-core/emscripten/issues/13025 that got a reply didn't get a response that curl doesn't work, just that sockets is very limited. Even going away from curl xeus-cpp could make use of Emscriptens fetch API https://emscripten.org/docs/api_reference/fetch.html .

Hi, 

Let's not stress too much on this. Not sure what you mean by completely ignore because I've just disabled it for the emscripten build cause currently we don't have a way to get this work.

Anyways to get any AI support in the browser, I think we would just go the [JupyterliteAI](https://github.com/jupyterlite/ai) route when that's ready.

anutosh491:
Hey @vgvassilev @JohanMabille 

I think this should be ready now. Apart from the 2 tests that have been added as TODOs rest all work (45/47) so really happy with the results. This should be good to get us started I think.

mcbarton:
> Hey @vgvassilev @JohanMabille
> 
> I think this should be ready now. Apart from the 2 tests that have been added as TODOs rest all work (45/47) so really happy with the results. This should be good to get us started I think.

@vgvassilev @JohanMabille This comment should be addressed before any approval is given https://github.com/compiler-research/xeus-cpp/pull/277#discussion_r1999477200 . I don't believe xeus-cpp should rely on Emscripten forge setting LD_FLAGS. Where possible in the build process we should maintain compatibility with the upstream emsdk.

anutosh491:
> I don't believe xeus-cpp should rely on Emscripten forge setting LD_FLAGS

Hi
Once the move to emscripten 3.1.73 all xeus lite kernels currently stick to use emsdk from emscripten-forge where `WASM_BIGINT` is a default link flag along with some others.

> Where possible in the build process we should maintain compatibility with the upstream emsdk.

I don't think xeus-cpp or any other xeus-lite kernel (or for that matter any Jupyterlite kernel) has ever claimed this. I don't think this is a hard necessity.

anutosh491:
I've made the change to keep everyone happy ! 

Should be good to go now.

anutosh491:
Hey @vgvassilev this should be ready now. Could you please have a look and merge if you're happy ?

PR #278, opened 2025-03-08
Added tests for xinspect

### LINKED ISSUE #276: Improve test coverage for xinspect.cpp
These set of lines stand untested right now (go through comment below)

https://github.com/compiler-research/xeus-cpp/pull/269#issuecomment-2703264890

Our goal is to have as much coverage for tests as possible. Would be great if this can be addressed by adding tests in https://github.com/compiler-research/xeus-cpp/blob/main/test%2Ftest_interpreter.cpp

kr-2003:
# Description

- Added tests for ```fetch_documentation_of_member_or_parameter``` and ```is_inspect_request```.
- Fixed ```inspect``` function. (Previously, one can't fetch documentation of member functions and variables.)

Fixes #276

### DIFF
--- a/test/test_interpreter.cpp
+++ b/test/test_interpreter.cpp
@@ -134,6 +134,43 @@ TEST_SUITE("execute_request")
         REQUIRE(result["status"] == "ok");
     }
 
+    TEST_CASE("fetch_documentation_of_member_or_parameter")
+    {
+        std::vector<const char*> Args = {/*"-v", "resource-dir", "....."*/};
+        xcpp::interpreter interpreter((int)Args.size(), Args.data());
+
+        std::string code = "?std::vector.push_back";
+        std::string inspect_result = "https://en.cppreference.com/w/cpp/container/vector/push_back";
+        nl::json user_expressions = nl::json::object();
+        xeus::execute_request_config config;
+        config.silent = false;
+        config.store_history = false;
+        config.allow_stdin = false;
+        nl::json header = nl::json::object();
+        xeus::xrequest_context::guid_list id = {};
+        xeus::xrequest_context context(header, id);
+
+        std::promise<nl::json> promise;
+        std::future<nl::json> future = promise.get_future();
+        auto callback = [&promise](nl::json result) {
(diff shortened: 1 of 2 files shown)


test/test_interpreter.cpp
```diff
@@ -840,6 +877,20 @@ TEST_SUITE("xinspect"){
         cmp.child_value = "nonexistentMethod";
         REQUIRE(cmp(node) == false);
     }
+
+    // TEST_CASE("find_type_slow"){
```
kr-2003, line 881: some issue with this test passing in ubuntu machines.

kr-2003:
@anutosh491 
We discussed about ``?std::vector.push_back``` working but ```?vector.push_back``` not working. And ```?std::vector``` and ```?vector``` working. 

This is not something that I can change. The format of ```cppreference-doxygen-web.tag``` file is such that it has 2 instances for vector. One is ```std::vector``` as class and another one is ```vector``` as file. Therefore one can find documentation related to both of them. But for any class member function or variable, it can be only access via class ```std::vector```. Therefore the above specified scenario takes place.

I think this is how it is supposed to work. 

PR is complete from my side.

anutosh491 (approved):
LGTM !

Coverage goes up. Works with lite too so no complaints. Thanks.

test/test_interpreter.cpp
```diff
@@ -134,6 +134,43 @@ TEST_SUITE("execute_request")
         REQUIRE(result["status"] == "ok");
     }
 
+    TEST_CASE("fetch_documentation_of_member_or_parameter")
+    {
+        std::vector<const char*> Args = {/*"-v", "resource-dir", "....."*/};
+        xcpp::interpreter interpreter((int)Args.size(), Args.data());
+
+        std::string code = "?std::vector.push_back";
+        std::string inspect_result = "https://en.cppreference.com/w/cpp/container/vector/push_back";
```
vgvassilev, line 143: Do we require internet connection for this test to work?

kr-2003: yes

PR #280, opened 2025-03-18
Make use of EMSDK environment variable which comes from sourcing emsdk script

mcbarton:
# Description

xeus-cpp instructions create a BUILD_PREFIX and SYSROOT_PATH environment variable. It would be better to just utilise the EMSDK environment variable which comes from when you setup emsdk. Also since we only use SYSROOT_PATH once it makes no sense for it to be an environment variable. We can just put the path directly into the cmake configure stage. 

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -50,17 +50,15 @@ jobs:
           mkdir build
           pushd build
 
-          export BUILD_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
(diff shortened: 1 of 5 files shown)


mcbarton:
The fact the EMSDK environment variable is blank when you get emsdk from Emscripten forge suggests a bug in the Emscripten Forge to me. You get this variable when you setup emsdk using the upstream repo (see https://github.com/compiler-research/CppInterOp/actions/runs/13933008473/job/38994499405#step:9:103) .

PR #288, opened 2025-04-16
Disable xsystem for WASM

tharun571:
# Description

Shell commands do not work in WASM environment. This PR disables it.

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/src/xsystem.hpp
+++ b/src/xsystem.hpp
@@ -28,6 +28,14 @@ namespace xcpp
 
         void apply(const std::string& code, nl::json& kernel_res) override
         {
+#if defined(XEUS_CPP_EMSCRIPTEN_WASM_BUILD)
+            // WASM environment: Disable shell commands
+            kernel_res["status"] = "error";
+            kernel_res["ename"] = "UnsupportedEnvironment";
+            kernel_res["evalue"] = "Shell commands are not supported in the WASM environment.";
+            kernel_res["traceback"] = nl::json::array();
+#else
+            // Native environment: Execute shell commands
             std::regex re(spattern + R"((.*))");
             std::smatch to_execute;
             std::regex_search(code, to_execute, re);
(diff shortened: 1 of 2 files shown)


test/test_interpreter.cpp
```diff
@@ -657,12 +657,20 @@ TEST_SUITE("xsystem_clone")
...
+        xcpp::xsystem system;
+        std::string code = "!echo Hello, World!";
+        nl::json kernel_res;
+
+        system.apply(code, kernel_res);
+
+        REQUIRE(kernel_res["status"] == "error");
+        REQUIRE(kernel_res["ename"] == "UnsupportedEnvironment");
+        REQUIRE(kernel_res["evalue"] == "Shell commands are not supported in the WASM environment.");
+    }
```
anutosh491, line 671: Hey Tharun,

let's just skip it for now ! Cause technically its not an error on the kernel status side correct ? 
Its just not expected to work. So technically our TODO : "Currently fails for the Emscripten build" isn't justified.

Cause it might fail but its not supposed to work either. let's just skip this test entirely !

tharun571: Makes sense.

src/xsystem.hpp
```diff
@@ -28,6 +28,14 @@ namespace xcpp
...
         void apply(const std::string& code, nl::json& kernel_res) override
         {
+#if defined(XEUS_CPP_EMSCRIPTEN_WASM_BUILD)
+            // WASM environment: Disable shell commands
+            kernel_res["status"] = "error";
+            kernel_res["ename"] = "UnsupportedEnvironment";
+            kernel_res["evalue"] = "Shell commands are not supported in the WASM environment.";
+            kernel_res["traceback"] = nl::json::array();
+#else
+            // Native environment: Execute shell commands
```
anutosh491, line 38: These changes won't be needed anymore !

anutosh491 (approved):
Looks good. 
This is not expected to work for the wasm build (so its not that it fails, it just isn't expected to work .... Can we make that a bit more clear in the PR description ?) !!!

Issue #222, opened 2025-01-15
Fix exception handling (throw) in xeus-cpp-lite.


anutosh491:
I would expect this to work with what is there as of now but I see the following doesn't work 

<img width="838" alt="image" src="https://github.com/user-attachments/assets/89566514-f573-4233-ab3e-c4dea60f89b0" />

1) The [linking step](https://github.com/llvm/llvm-project/blob/2a044f8a092efb27fa1837f953bce8237d41e59b/clang/lib/Interpreter/Wasm.cpp#L113) through clang-repl is successful here and hence a shared wasm module has been created.

2) But the dlopen step to load it on top of the main module is not working as expected. Hence the trace

<img width="1097" alt="image" src="https://github.com/user-attachments/assets/f5f8a87e-2de3-4c32-84d9-5b7ea35a52d1" />

anutosh491:
The trace indicates that the `loadDynamicLibrary` function call fails from dlopenInternal leading to the error 
```
        var dlopenInternal = (handle, jsflags) => {
            // void *dlopen(const char *file, int mode);
            // http://pubs.opengroup.org/onlinepubs/009695399/functions/dlopen.html
            var filename = UTF8ToString(handle + 36);
            var flags = HEAP32[(((handle) + (4)) >> 2)];
            filename = PATH.normalize(filename);
            var searchpaths = [];

            var global = Boolean(flags & 256);
            var localScope = global ? null : {};

            // We don't care about RTLD_NOW and RTLD_LAZY.
            var combinedFlags = {
                global,
                nodelete: Boolean(flags & 4096),
                loadAsync: jsflags.loadAsync,
            }

            if (jsflags.loadAsync) {
                return loadDynamicLibrary(filename, combinedFlags, localScope, handle);
            }

            try { // THIS TRY BLOCK FAILS !!!!
                return loadDynamicLibrary(filename, combinedFlags, localScope, handle)
            } catch (e) {
                err(`Error in loading dynamic library ${filename}: ${e}`);
                dlSetError(`Could not load dynamic lib: ${filename}\n${e}`);
                return 0;
            }
        }
        
```

anutosh491:
Also this appears in the console 

<img width="1093" alt="image" src="https://github.com/user-attachments/assets/fd2a252b-9d12-4dbb-81e1-d4d7fd8c6bdd" />

anutosh491:
[...]

PR #290, opened 2025-04-17
Enable support for multiple kernels for xeus-cpp-lite

anutosh491:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,7 @@ function(configure_native_kernel kernel)
 endfunction()
 
 function(configure_wasm_kernel kernel)
+  set(XEUS_CPP_RESOURCE_DIR "/lib/clang/${CPPINTEROP_LLVM_VERSION_MAJOR}")
 
   configure_file (
     "${CMAKE_CURRENT_SOURCE_DIR}${kernel}wasm_kernel.json.in"
@@ -169,17 +170,9 @@ endfunction()
 
 message("Configure kernels: ...")
 if(EMSCRIPTEN)
-    # TODO: Currently jupyterlite-xeus and xeus-lite do not provide
-    # methods to fetch information from the arguments present in the
-    # generated emscripten kernel.
-    # The following needs to be done here :
-    # 1) We need to configure the kernel properly
-    #    Check issue https://github.com/compiler-research/xeus-cpp/issues/185.
-    # 2) Once the above is done we need to add support in jupyterlite-xeus & xeus-lite
-    #    to be able to deal with arguments present in kernel.json
-    # 3) Finally we should fetch the C++ version from the kernel.json file and
-    #    be able to pass it to our wasm interpreter rather than forcing a version.
+    configure_wasm_kernel("/share/jupyter/kernels/xcpp17/")
     configure_wasm_kernel("/share/jupyter/kernels/xcpp20/")
+    configure_wasm_kernel("/share/jupyter/kernels/xcpp23/")
 else()
     configure_native_kernel("/share/jupyter/kernels/xcpp17/")
     configure_native_kernel("/share/jupyter/kernels/xcpp20/")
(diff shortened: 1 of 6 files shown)


anutosh491:
https://github.com/user-attachments/assets/d6dc6d5f-6f02-4618-bb02-34543741765a

Demo above : Shows 3 xeus-cpp-lite kernels built (all use different flags -std=c++17/20/23)

On top of which xcpp20 is built using the -msimd128 flag (required for running simd code in the browser)

So xcpp20 kernel would work but xcpp23 wouldn't !

CMakeLists.txt
```diff
@@ -444,20 +437,13 @@ if(EMSCRIPTEN)
     xeus_wasm_link_options(xcpp "web,worker")
     target_link_options(xcpp
         PUBLIC "SHELL: --preload-file ${SYSROOT_PATH}/include@/include"
+        # PUBLIC "SHELL: --preload-file ${CMAKE_INSTALL_PREFIX}${XEUS_CPP_RESOURCE_DIR}/include@${XEUS_CPP_RESOURCE_DIR}"
```
mcbarton, line 440: I can't work out what this commented out line has got to do with the objective of this PR.  I see the below
```
    # TODO: Uncomment the above line regarding preloading clang's resource dir
    # once has been supported through cppinterop.
```
Seems unrelated to getting multiple kernels working, but maybe i'm missing something.

anutosh491: Well all kernels would be needing resource dir support in the future correct ? 
This makes sure we have that ..... whenever we do ! (we would just need to uncomment this line out)

So we keep our kernel.json ready to deal with resource dirs whenever we are ready.

Anyways this PR needs fixing things on xeus-lite and jupyterlite first, followed by a release. So yeah I don't think there's any value discussing this right now.

anutosh491:
clang-format is a bit too much (probably can be ignored here ?)

src/xinterpreter.cpp
```diff
@@ -27,10 +27,7 @@
 using Args = std::vector<const char*>;
 
 void* createInterpreter(const Args &ExtraArgs = {}) {
-  Args ClangArgs = {/*"-xc++"*/"-v"}; // ? {"-Xclang", "-emit-llvm-only", "-Xclang", "-diagnostic-log-file", "-Xclang", "-", "-xc++"};
-#ifdef EMSCRIPTEN
-  ClangArgs.push_back("-std=c++20");
-#else
```
anutosh491, line 33: We don't need to hardcode anything for xeus-cpp-lite from now on. The user can vary the kernel.json as per how he likes.

anutosh491:
Hey @JohanMabille @martinRenou

As y'all helped me with the design here on the JupyterLite, could y'all review the changes here for me ?

1) After this we should see 3 kernels through our deployment link
2) We are preloading the path to clang's resource-dir (/lib/clang/20 where 20 is latest llvm version) into the file system. This provides the header wasm_simd128.h for xsimd to run in the browser.
3) As a ToDo 
After the next cppinterop release, we would also be able to address this https://github.com/compiler-research/xeus-cpp/issues/222

So currently xeus-cpp-lite doesn't have exception handling due to missing emscripten flags. 
But we should soon be able to make this change 
```
{
  "display_name": "C++20",
  "argv": [
      "@XEUS_CPP_KERNELSPEC_PATH@xcpp",
      "-resource-dir", "@XEUS_CPP_RESOURCE_DIR@",
-      "-std=c++20"
+      "-std=c++20", "-mllvm", "-enable-emscripten-cxx-exceptions"
  ],
  
 ```
 And support this as I tried building on main
 
image

**Review by martinRenou — APPROVED** [review]
Really cool to finally get multiple kernel specs support in JupyterLite!

**Review by JohanMabille — CHANGES_REQUESTED** [review]

File: src/main_emscripten_kernel.cpp [hunk 3]
```diff
@@ -15,9 +15,26 @@
...
+
+    for (const auto& s : args)
+    {
+        argv_vec.push_back(s.c_str());
+    }
+
+    int argc = static_cast<int>(argv_vec.size());
+    char** argv = const_cast<char**>(argv_vec.data());
+
+    return new interpreter_type(argc, argv);
```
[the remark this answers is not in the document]
anutosh491: Ahh yes, thanks for pointing this out.

I did consider changing the interpreter's constructor here (but thought of possibly doing that through a separate PR) and was also considering a similar line of thought for `args`/`arg_vec`

But then I thought the builder is intended to be used **only once** to construct the interpreter (not sure if we would have a case where we call builder_with_args multiple times in the near future ig ?) 

Hence I went ahead with the `static args/argv_vec` approach.

And yes as correctly pointed out the interpreter makes an internal copy of argv data during
construction, so we do not need to retain ownership of the vector afterward and can clear them.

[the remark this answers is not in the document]
anutosh491: Addressed this in the latest commit (added a small TODO to improve the interpreter constructor)

JohanMabille (approved):

Issue #249, opened 2025-01-30
std::cin fails with xeus-cpp-lite


anutosh491:
On Xeus-cpp locally

<img width="827" alt="Image" src="https://github.com/user-attachments/assets/21feb042-1fcc-4914-bbd2-5058189cf184" />

**On Xeus-cpp-lite (can be tried through the deployment here https://compiler-research.org/xeus-cpp/lab/index.html)**

<img width="827" alt="Image" src="https://github.com/user-attachments/assets/ba844d19-e154-4597-a00c-787497afb583" />

anutosh491:
Should be addressed as soon as Jupyterlite 0.6.0 is released.

Check https://github.com/jupyterlite/jupyterlite/pull/1640

Not sure if there's a date attached to this milestone !?

PR #291, opened 2025-04-22
Pin nlohmann_json to 3.12.0

mcbarton:
# Description

The latest release of xeus got released, and requires nlohmann_json to be 3.12.0 . Until this PR is in the ci will fail
(see https://github.com/compiler-research/xeus-cpp/actions/runs/14604191183/job/40969335106#step:5:97).

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -3,9 +3,9 @@ channels:
   - https://repo.prefix.dev/emscripten-forge-dev
   - https://repo.mamba.pm/conda-forge
 dependencies:
-  - nlohmann_json=3.11.3
+  - nlohmann_json
   - xeus-lite
-  - xeus=5.2.0
+  - xeus
   - CppInterOp
   - cpp-argparse
   - pugixml
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -3,7 +3,7 @@ channels:
   - https://repo.prefix.dev/emscripten-forge-dev
   - https://repo.mamba.pm/conda-forge
 dependencies:
-  - nlohmann_json
+  - nlohmann_json=3.12.0
   - xeus-lite
   - xeus
   - CppInterOp
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -3,9 +3,9 @@ channels:
   - https://repo.prefix.dev/emscripten-forge-dev
   - https://repo.mamba.pm/conda-forge
 dependencies:
-  - nlohmann_json
+  - nlohmann_json=3.11.3
   - xeus-lite
-  - xeus
+  - xeus=5.2.0
   - CppInterOp
   - cpp-argparse
   - pugixml
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -3,7 +3,7 @@ channels:
   - https://repo.prefix.dev/emscripten-forge-dev
   - https://repo.mamba.pm/conda-forge
 dependencies:
-  - nlohmann_json=3.11.3
+  - nlohmann_json
   - xeus-lite
   - xeus
   - CppInterOp
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -9,7 +9,7 @@ dependencies:
   # Host dependencies
   - xeus>=5.0.0
   - xeus-zmq<4.0
-  - nlohmann_json=3.11.3
+  - nlohmann_json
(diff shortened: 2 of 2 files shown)


environment-wasm-host.yml
```diff
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - nlohmann_json=3.11.3
   - xeus-lite
-  - xeus
+  - xeus=5.2.0
```
anutosh491, line 8: I am not sure pinning xeus makes sense. Why should the native kernels use the latest version but the lite kernels use an older version.

Also the update in xeus is related to fixing this issue for lite kernels
https://github.com/compiler-research/xeus-cpp/issues/249 !

anutosh491:
cc @JohanMabille this is actually related to what I was mentioning yesterday

Here the keyword REQUIRED is put to use (so building xeus would need anything above 3.11)
https://github.com/jupyter-xeus/xeus/blob/7e27db78137ab8678de28a06e70af1996e334e40/CMakeLists.txt#L78-L88

and here the keyword EXACT is put to use (building downstream projects would need exactly 3.12)
https://github.com/jupyter-xeus/xeus/blob/7e27db78137ab8678de28a06e70af1996e334e40/xeusConfig.cmake.in#L25C1-L27C61

Does this sound like a mismatch to you ?

anutosh491:
I think we just need to remove the pins overall 

Check : https://github.com/emscripten-forge/recipes/pull/2270

mcbarton:
> I think we just need to remove the pins overall
> 
> Check : [emscripten-forge/recipes#2270](https://github.com/emscripten-forge/recipes/pull/2270)
> 
> Just remove both pins and we should be good

@anutosh491 I already tried this and it didn't work (see https://github.com/compiler-research/xeus-cpp/actions/runs/14604401760/job/40970026439?pr=291 ). That is the reason I pinned xeus as a temporary fix until the error with everything unpinned could be resolved.

anutosh491:
There are differences! 

Your commit uses xeus-lite 3.0.1 and the latest is 3.1.0 

Please remove the pins and build again and we should be good I think

mcbarton:
> There are differences!
> 
> Your commit uses xeus-lite 3.0.1 and the latest is 3.1.0
> 
> Please remove the pins and build again and we should be good I think !
> 
> Change the PR heading accordingly

@anutosh491 Done. I've also done the same to my PR in CppInterOp here https://github.com/compiler-research/CppInterOp/pull/566 . Please approve both at the same time

anutosh491:
Your pr still uses xeus 5.2.0 for the native case. 

We technically need to unpin this too https://github.com/compiler-research/xeus-cpp/blob/main/environment-dev.yml#L12

which hasn't been done.

anutosh491:
Not sure why are we still fetching 5.2.0  :(

image

JohanMabille:
> cc @JohanMabille this is actually related to what I was mentioning yesterday
> 
> Here the keyword REQUIRED is put to use (so building xeus would need anything above 3.11) https://github.com/jupyter-xeus/xeus/blob/7e27db78137ab8678de28a06e70af1996e334e40/CMakeLists.txt#L78-L88
> 
> and here the keyword EXACT is put to use (building downstream projects would need exactly 3.12) https://github.com/jupyter-xeus/xeus/blob/7e27db78137ab8678de28a06e70af1996e334e40/xeusConfig.cmake.in#L25C1-L27C61
> 
> Does this sound like a mismatch to you ?

It is not a mismatch. Xeus can work with any version compatible with 3.11, so we require 3.11. Any downstream project that will exchange json objects with xeus needs to be built with the same version as the one used to build xeus for ABI compatibility reasons.

The pinning on nlohmann_json ensures this; otherwise, you may endup downloading a newer version of nolhmann_json from conda-forge and notive the failure at build time only. Therefore, this pinning should not be removed, but updated to that used to build xeus.

A better solution would be a global pinning on conda-forge, but we don't have it for now.

mcbarton:
> > Therefore, this pinning should not be removed, but updated to that used to build xeus.
> 
> Yes this might be the way forward. Let's pin it to 3.12.0 for now

Done.

anutosh491:
> Done.

Ahh I think the error calls for building xeus-zmq with the latest xeus which hasn't been done. Looking into it !

anutosh491 (approved):
Now ready. Thanks !

PR #292, opened 2025-04-27
Removed build_interpreter from main()

kr-2003:
# Description

In the main function, call to ```build_interpreter``` is redundant. Instead calling directly to ```xcpp::interpreter::interpreter``` will work fine. 

### Reasoning

```cpp
interpreter_ptr build_interpreter(int argc, char** argv)
{
    std::vector<const char*> interpreter_argv(argc);
    interpreter_argv[0] = "xeus-cpp";

    // Copy all arguments in the new vector excepting the process name.
    for (int i = 1; i < argc; i++)
    {
        interpreter_argv[i] = argv[i];
    }

    interpreter_ptr interp_ptr = std::make_unique<interpreter>(argc, interpreter_argv.data());
    return interp_ptr;
}
```
This sets first arg and then passes all the args to ```xcpp::interpreter::interpreter```. But ```xcpp::interpreter::interpreter``` always skips the first arg(if the argv is non-null). 
```cpp
interpreter::interpreter(int argc, const char* const* argv) :
    xmagics()
    , p_cout_strbuf(nullptr)
    , p_cerr_strbuf(nullptr)
    , m_cout_buffer(std::bind(&interpreter::publish_stdout, this, _1))
    , m_cerr_buffer(std::bind(&interpreter::publish_stderr, this, _1))
{
    //NOLINTNEXTLINE (cppcoreguidelines-pro-bounds-pointer-arithmetic)
    createInterpreter(Args(argv ? argv + 1 : argv, argv + argc));
    m_version = get_stdopt();
    redirect_output();
    init_preamble();
    init_magic();
}
```

But, what if the argv is ``nullptr``??. It can happen if the user manually removes all the args from ``kernel.json``(including the program name). 
It can be show that if ``we are not passing any argument`` is same as ``we are passing only one argument``(i.e. the pgroam name). 
``createInterpreter(Args(nullptr, nullptr));`` is same as ``createInterpreter(Args(argv + 1, argv + 1));``.

#### Proof

Screenshot 2025-04-27 at 12 42 53 PM

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -59,7 +59,7 @@ int main(int argc, char* argv[])
     signal(SIGINT, xcpp::stop_handler);
 
     std::string file_name = xeus::extract_filename(argc, argv);
-    interpreter_ptr interpreter = xcpp::build_interpreter(argc, argv);
+    auto interpreter = std::unique_ptr<xcpp::interpreter>(new xcpp::interpreter(argc, argv));
     std::unique_ptr<xeus::xcontext> context = xeus::make_zmq_context();
 
     if (!file_name.empty())
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -59,7 +59,7 @@ int main(int argc, char* argv[])
     signal(SIGINT, xcpp::stop_handler);
 
     std::string file_name = xeus::extract_filename(argc, argv);
-    auto interpreter = std::unique_ptr<xcpp::interpreter>(new xcpp::interpreter(argc, argv));
+    auto interpreter = std::make_unique<xcpp::interpreter>(argc, argv);
     std::unique_ptr<xeus::xcontext> context = xeus::make_zmq_context();
 
     if (!file_name.empty())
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -59,7 +59,7 @@ int main(int argc, char* argv[])
     signal(SIGINT, xcpp::stop_handler);
 
     std::string file_name = xeus::extract_filename(argc, argv);
-    interpreter_ptr interpreter = xcpp::build_interpreter(argc, argv);
+    auto interpreter = std::unique_ptr<xcpp::interpreter>(new xcpp::interpreter(argc, argv));
```
anutosh491, line 62: Probably 
```
auto interpreter = std::unique_ptr< interpreter_ptr >(new interpreter_ptr(argc, argv));
```

Looks better isn't it ?

kr-2003: i guess the one recommended by clang-tidy is much more better??
```
 auto interpreter = std::make_unique<xcpp::interpreter>(argc, argv);
```

anutosh491 (approved):
As I was going through the creation of the interpreter, i realized we anyways skip the first argument anways 

https://github.com/compiler-research/xeus-cpp/blob/1158ecd728479c2386320cb118123f252b23b77b/src/xinterpreter.cpp#L112

So having 1 arg in the container is equally good as nothing. 

This makes build_interpreter redundant as we just make some change on the 0th element through it !

PR #293, opened 2025-05-01
Added undo operation for xeus-cpp

### LINKED ISSUE #263: Handle undo for xeus-cpp/xeus-cpp-lite
Clang-repl provides an undo option which I guess should help us get past redefinition issues .
https://llvm.org/devmtg/2022-11/slides/Lightning6-RecoveringFromErrorsInClang-ReplAndCodeUndo.pdf

For example 
```
anutosh491@Anutoshs-MacBook-Air bin % ./clang-repl
clang-repl> extern "C" int printf(const char*, ...);
clang-repl> int x = 50;
clang-repl> %undo
clang-repl> const char* x = "Hello World";
clang-repl> auto r = printf("%s\n", x);
Hello World
```

Without the undo you would get 
```
clang-repl> int y = 50;
clang-repl> const char* y = "Hello World";
In file included from <<< inputs >>>:1:
input_line_6:1:13: error: redefinition of 'y' with a different type: 'const char *' vs 'int'
    1 | const char* y = "Hello World";
      |             ^
input_line_5:1:5: note: previous definition is here
    1 | int y = 50;
      |     ^
error: Parsing failed.
```

When the same is attempted through xeus-cpp, it is being considered as a magic but probably should be looked at from a simple execute_request point of view .

Image

kr-2003:
## Description

Undo last N operations.

Fixes #263

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -134,14 +134,17 @@ __get_cxx_version ()
 
         auto input_guard = input_redirection(config.allow_stdin);
 
-        // Check for magics
-        for (auto& pre : preamble_manager.preamble)
+        if (code.find("%undo") != 0)
         {
-            if (pre.second.is_match(code))
+            // Check for magics
+            for (auto& pre : preamble_manager.preamble)
             {
-                pre.second.apply(code, kernel_res);
-                cb(kernel_res);
-                return;
+                if (pre.second.is_match(code))
+                {
+                    pre.second.apply(code, kernel_res);
+                    cb(kernel_res);
+                    return;
+                }
             }
         }
 
@@ -169,7 +172,32 @@ __get_cxx_version ()
         try
         {
             StreamRedirectRAII R(err);
-            compilation_result = Cpp::Process(code.c_str());
+            if (code.rfind("%undo", 0) == 0)
+            {
+                int n = 1;  // Default value
+                if (code.length() > 5)
+                {
+                    try
+                    {
+                        n = std::stoi(code.substr(6));
+                    }
+                    catch (const std::invalid_argument&)
+                    {
+                        throw std::runtime_error(
+                            "Invalid format for %undo. Expected '%undo n' where n is an integer."
+                        );
+                    }
+                    catch (const std::out_of_range&)
+                    {
+                        throw std::runtime_error("Number out of range for %undo.");
+                    }
+                }
+                compilation_result = Cpp::Undo(n) ? true : false;
+            }
+            else
+            {
+                compilation_result = Cpp::Process(code.c_str());
+            }
         }
         catch (std::exception& e)
         {


anutosh491:
Maybe you can add a couple tests ?

kr-2003:
> Maybe you can add a couple tests ?

Sure.

anutosh491:
Hey @kr-2003, 

I see lot of formatting changes and stuff here. Let's keep implementation prs and refactoring/formatting prs separately. 

I know sometimes the ci would suggest some edits, but that's just overhead. We shouldn't necessarily respect it everytime as it just pollutes commit history.

Could you please clean this up to just have the core changes required here to enable undo ?

mcbarton:
Fascinating that the Emscripten tests pass for the MacOS build, but not the Linux Emscripten build here https://github.com/compiler-research/xeus-cpp/actions/runs/15210235224

mcbarton:
@kr-2003 can you add a cell to the example notebook we use for native builds, showing the undo feature in action?

src/xinterpreter.cpp
```diff
@@ -168,7 +171,42 @@ __get_cxx_version ()
         try
         {
             StreamRedirectRAII R(err);
-            compilation_result = Cpp::Process(code.c_str());
+            if (code.rfind("%undo", 0) == 0)
+            {
+#ifdef EMSCRIPTEN
+                throw std::logic_error("Undo is not supported in xeus-cpp-lite");
```
mcbarton, line 177: This PR is linked to closing the issue 'Handle undo for xeus-cpp/xeus-cpp-lite' but this PR specifically mentions that undo is not supported for xeus-cpp lite

kr-2003: Since, undo for lite will take some time to implement, this PR will not close the issue. Edited the PR description.

kr-2003:
> @kr-2003 can you add to the example notebook we use for native builds, showing the undo feature in action?

Sure, I’ll add it in the meantime

anutosh491:
> We might need to rethink the entire meaning of undo in the context of jupyter notebooks.

Hmm, I thought we would just stick to what clang-repl/cppinterop was doing. 

"%undo" looks like an interesting trick though haha ( so I thought we could introduce it as a magic ... %%undo I guess ?)

kr-2003:
> We might need to rethink the entire meaning of `undo` in the context of jupyter notebooks. We probably want to change the granularity of the undo operations to be on cell level. In that case, however, we will need some level of support to "grey-out" the undone cells...

Just to clarify — when you mention rethinking the meaning of undo and changing the granularity to the cell level, do you mean that using %undo should erase all the statements from the previous cell?

If so, it seems like %undo already handles that behavior, as shown here:
undo

In that case, the only remaining enhancement might be adding support for visually greying-out the undone cells.

PR #294, opened 2025-05-01
Run Emscripten tests in a browser

mcbarton:
# Description

This PR adds the ability to run the Emscripten tests in a browser

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -51,6 +51,7 @@ jobs:
           pushd build
 
           export BUILD_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
+          echo "BUILD_PREFIX=$BUILD_PREFIX" >> $GITHUB_ENV
           export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-host 
           echo "PREFIX=$PREFIX" >> $GITHUB_ENV
           export SYSROOT_PATH=$BUILD_PREFIX/opt/emsdk/upstream/emscripten/cache/sysroot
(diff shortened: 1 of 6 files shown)


test/CMakeLists.txt
```diff
@@ -69,6 +71,8 @@ if(EMSCRIPTEN)
         PUBLIC "SHELL: --preload-file ${SYSROOT_PATH}/include@/include"
         PUBLIC "SHELL: --preload-file ../${XEUS_CPP_DATA_DIR}@/share/xeus-cpp"
         PUBLIC "SHELL: --preload-file ../${XEUS_CPP_CONF_DIR}@/etc/xeus-cpp"
+        PUBLIC "SHELL: --emrun"
+        PUBLIC "SHELL: -s MINIFY_HTML=0"
```
mcbarton, line 75: Without this MINIFY_HTML=0 flag you end up with the situation where the creation of the test_xeus_cpp.html file breaks on MacOS, but not Ubuntu as can be seen here https://github.com/compiler-research/xeus-cpp/actions/runs/14780201517

mcbarton:
Running the tests makes use of emrun which comes with Emscripten. The documentation for Emscripten can be found here https://emscripten.org/docs/compiling/Running-html-files-with-emrun.html

.github/workflows/deploy-github-page.yml
```diff
@@ -64,6 +65,60 @@ jobs:
...
 
+      - name: Test xeus-cpp C++ Emscripten
+        shell: bash -l {0}
+        run: |
+          set -e
+          micromamba create -n xeus-lite-host jupyterlite-core
+          micromamba activate xeus-lite-host
+          cd build/test
+          os="${{ matrix.os }}"
+          if [[ "${os}" == "macos"* ]]; then
```
mcbarton, line 76: The reason we install a fresh install of Firefox and Chome is to follow what is done in the Emscripten repo (for example here with Firefox https://github.com/emscripten-core/emscripten/blob/38e16464cffc8f886364fe4919a712131a4c9456/.circleci/config.yml#L363 ). It has the added advantage by not using the users installed versions of these browsers (if they have them installed), that there are no add ons, etc that could be causing issues with the tests running. I have made the instructions, such that a developer without admin privileges on there machine can install them.

.github/workflows/deploy-github-page.yml
```diff
@@ -64,6 +65,60 @@ jobs:
...
+            # Install Firefox
+            wget https://ftp.mozilla.org/pub/firefox/releases/138.0.1/linux-x86_64/en-GB/firefox-138.0.1.tar.xz
+            tar -xJf firefox-138.0.1.tar.xz
+            cd ./firefox
+            export PATH="$PWD:$PATH"
+            cd -
+
+            # Run tests in browsers
+            echo "Running test_xeus_cpp in Firefox"
+            python ${{ env.BUILD_PREFIX }}/bin/emrun.py --browser="firefox" --kill_exit --timeout 60 --browser-args="--headless"  test_xeus_cpp.html
```
mcbarton, line 115: Given emrun is not on our PATH, I have to reference the exact path to emrun.py . I could add `${{ env.BUILD_PREFIX }}/bin/` to PATH, but not sure if this would be wanted.

`--kill_exit` makes it so that when emrun finishes, that the headless browser we create is killed along with it
`--timeout 60` is such that emrun is killed after 60 seconds if still running. emrun should have finished long before then, so if it is still running, something went wrong (such as a test which crashed the html file). This will cause the ci to fail, as a non 0 value of will be returned.

.github/workflows/deploy-github-page.yml
```diff
@@ -64,6 +65,60 @@ jobs:
...
+            echo "Running test_xeus_cpp in Google Chrome"
+            python ${{ env.BUILD_PREFIX }}/bin/emrun.py --browser="google-chrome" --kill_exit --timeout 60 --browser-args="--headless  --no-sandbox"  test_xeus_cpp.html
+          fi
```
mcbarton, line 118: In the case of Chrome we have the extra `--no-sandbox` flag, as on Ubuntu Chrome will refuse to run otherwise, as it expects to have been installed with admin privileges. This flag allows it to run in userspace.

test/CMakeLists.txt
```diff
@@ -69,6 +71,8 @@ if(EMSCRIPTEN)
         PUBLIC "SHELL: --preload-file ${SYSROOT_PATH}/include@/include"
         PUBLIC "SHELL: --preload-file ../${XEUS_CPP_DATA_DIR}@/share/xeus-cpp"
         PUBLIC "SHELL: --preload-file ../${XEUS_CPP_CONF_DIR}@/etc/xeus-cpp"
+        PUBLIC "SHELL: --emrun"
```
mcbarton, line 74: Rebuild your Emscripten application and add the --emrun [linker flag](https://emscripten.org/docs/tools_reference/emcc.html#emcc-emrun).

    This flag injects code into the generated [Module object](https://emscripten.org/docs/api_reference/module.html#module) to enable capture of stdout, stderr and exit().

    Note

    If you skip this step, you can still run any .html file with emrun, but the capture will not work.

The above is taken from emruns documentation https://emscripten.org/docs/compiling/Running-html-files-with-emrun.html

test/CMakeLists.txt
```diff
@@ -56,7 +56,9 @@ if(EMSCRIPTEN)
     target_compile_options(test_xeus_cpp
         PUBLIC "SHELL: -fexceptions"
     )
-
+    set_target_properties(test_xeus_cpp PROPERTIES
+        SUFFIX ".html"
+    )
```
mcbarton, line 61: So that we create a html file, as well as the javascript file.

test/CMakeLists.txt
vgvassilev: Can we capture these comments as in-source comments, too?

anutosh491:
Hey @DerThorsten

I would really appreciate your expertise here. Please let us if this should be our GoTo method to address the problem statement here.

The contributor has made quite some comments describing his approach here.

Feel free to let us know if we can do better. Thanks !

vgvassilev:
IIRC, @mcbarton mentioned he discussed that at the emscripten discord or similar.

mcbarton:
I've added the PR comments, as comments in the files as requested here https://github.com/compiler-research/xeus-cpp/pull/294#discussion_r2099424022

anutosh491:
Hey @SylvainCorlay ,

I went through this PR today and I don't have any strong opinions here (probably I'm not educated enough or falling short on time to educate myself with all going on here)

I tagged @DerThorsten couple days back but haven't heard from him. 

So do you have any opinions if this is how we should test stuff in the browser ? 
I'll let you move this in if you think we're good (Vassil says we're good)

Some surface level reviews 
1) Not a fan of having explanations/long writeups in the actions file. If at all I would like references/docs than explanation for some thing.  If in question we can get back to the PR as to why something was done. We all probably lack experiences that a technical writer would have, so framing explanations more than a couple lines is not really what I'd prefer.

2) Not a fan of having users follow a huge set of instructions in the readme for building/testing (1 configure command, 1 build command, 1 test command is all I would follow). If at all we want to provide a huge set of instructions for testing ....... I've seen quite some repos just put all of it in a build script and ask the user to run the build script (like ./build_xeus_cpp_test.sh) 

That being said, these are just my thoughts. If anyone thinks otherwise and can't wait on this, feel free to move it in.

SylvainCorlay:
For in-browser testing, my initial thought was that we could use Galata for full integration tests with Jupyter, including visual non-regression tests, but I like this approach of running the standard C++ tests in the browser with emrun.

Galata-based integration tests could be added at a later stage, including for the native builds.

PR #296, opened 2025-05-12
Add -s STRICT=1 flag emscripten build

mcbarton:
# Description

Taken from Emscripten documentation here https://emscripten.org/docs/tools_reference/settings_reference.html

```
Emscripten ‘strict’ build mode: Drop supporting any deprecated build options. Set the environment variable EMCC_STRICT=1 or pass -sSTRICT to test that a codebase builds nicely in forward compatible manner. Changes enabled by this:

        The C define EMSCRIPTEN is not defined (__EMSCRIPTEN__ always is, and is the correct thing to use).

        STRICT_JS is enabled.

        IGNORE_MISSING_MAIN is disabled.

        AUTO_JS_LIBRARIES is disabled.

        AUTO_NATIVE_LIBRARIES is disabled.

        DEFAULT_TO_CXX is disabled.

        USE_GLFW is set to 0 rather than 2 by default.

        ALLOW_UNIMPLEMENTED_SYSCALLS is disabled.

        INCOMING_MODULE_JS_API is set to empty by default.
```

Making sure the Emscripten build is done in such a way as to be forward compatible feels like a good feature to have.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,6 +441,8 @@ if(EMSCRIPTEN)
         PUBLIC "SHELL: --preload-file ${XEUS_CPP_DATA_DIR}@/share/xeus-cpp"
         PUBLIC "SHELL: --preload-file ${XEUS_CPP_CONF_DIR}@/etc/xeus-cpp"
         PUBLIC "SHELL: --post-js ${CMAKE_CURRENT_SOURCE_DIR}/wasm_patches/post.js"
+        PUBLIC "SHELL: -s STRICT=1"
+        PUBLIC "SHELL: --no-entry"
     )
     # TODO: Uncomment the above line regarding preloading clang's resource dir
     # once has been supported through cppinterop.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -441,6 +441,8 @@ if(EMSCRIPTEN)
         PUBLIC "SHELL: --preload-file ${XEUS_CPP_DATA_DIR}@/share/xeus-cpp"
         PUBLIC "SHELL: --preload-file ${XEUS_CPP_CONF_DIR}@/etc/xeus-cpp"
         PUBLIC "SHELL: --post-js ${CMAKE_CURRENT_SOURCE_DIR}/wasm_patches/post.js"
+        PUBLIC "SHELL: -s STRICT=1"
+        PUBLIC "SHELL: --no-entry"
```
mcbarton, line 445: This --no-entry flag is to deal with the error here https://github.com/compiler-research/xeus-cpp/pull/296/checks?sha=4042234c05975b3bb37749ddd2cb6be932e4a638#step:5:167 , which comes after we make our build have strict=1 (e.g. have forwards compatibility)

anutosh491:
Hi, 

Xeus-cpp's interpreter is based on xeus' xinterpreter, so we don't add flags directly here unless they are xeus-cpp specific (like the preloads). You should raise a PR for any generic flag here https://github.com/jupyter-xeus/xeus/blob/main/cmake/WasmBuildOptions.cmake#L21-L40 as these flags are propagated through this line https://github.com/compiler-research/xeus-cpp/blob/3c5328d1214e6531fbd1e396aa2c2795d3ff4bbb/CMakeLists.txt#L437

PR #235, opened 2025-01-23
Add smallpt as an example notebook

anutosh491:
# Description

Please include a summary of changes, motivation and context for this PR.

Replicating https://www.kevinbeason.com/smallpt/ through xeus-cpp-lite

## Type of change

Please tick all options which are relevant.

- [ ] Bug fix
- [x] New feature
- [ ] Added/removed dependencies
- [ ] Required documentation updates

PR #299, opened 2025-05-13
Add smallpt example notebook running through xeus-cpp-lite

anutosh491:
# Description
Refer https://github.com/compiler-research/xeus-cpp/pull/235#discussion_r2086283524

## Type of change

- [ ] Bug fix
- [x] New feature


anutosh491:
cc @kr-2003 , could you please verify the working of the smallpt example for me ?

It is 2 min process (so have add some logs so that the user doesn't feel lost)

You should be seeing this

image

kr-2003:
@anutosh491 
I have checked the smallpt notebook. Its working fine.

Screenshot 2025-05-13 at 5 50 22 PM

mcbarton:
Is there a reason this example is a completely separate notebook, and the examples is not added to the already existing notebook?

mcbarton:
If there is a legitimate reason to have it as a separate notebook, then it needs separate notebook needs to be uploaded to the deployment here https://github.com/anutosh491/xeus-cpp/blob/9628b1c132d23833b25ee08166c1ee2f71f3205e/.github/workflows/deploy-github-page.yml#L76

mcbarton (changes requested):
Example either needs integrating with existing notebook (if no legitimate reason to be separate), or new notebook added to deployment.

SylvainCorlay:
> the notebook needs to be uploaded to the deployment here

Good catch!

I think it makes sense for it to be in a separate notebook. The main notebook displays the basic features of the kernel, and this notebook shows a more advanced usecase.

anutosh491:
> Good catch!
>
> I think it makes sense for it to be in a separate notebook. The main notebook displays the basic features of the kernel, and this notebook shows a more advanced usecase.

Addressed !

SylvainCorlay:
This looks good to me! Thanks!

SylvainCorlay (approved):

SylvainCorlay:
@mcbarton it seems your comment has been addressed.

PR #312, opened 2025-05-25
Make LLVM_RESOURCE_DIR an option that can be provided by the user

mcbarton:
# Description

The llvm resource directory is now unconditionally assumed to be installed in the CMAKE_INSTALL_PREFIX . This has caused problems in CppInterOp where we do not install it in this location. Therefore this PR makes the llvm resource directory an option that can be provided by the user, and fallsback to assuming its in the install prefix if not defined.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature


mcbarton:
Will change this to be clang resource dir later today

PR #314, opened 2025-05-26
Clad plugin for xeus-cpp 

kr-2003:


vgvassilev:
> Hi,
> 
> Just trying to extrapolate this to the lite use case.
> 
> Can clad be built against emscripten -> and then used through -fplugin ?
> 
> I am not sure if clang plugins would work with lite (should they ? 🤔 )

If dlopen works then they should.

SylvainCorlay:
I think we need to hold off on including plugin without defining a plugin API surface, or in the core of the package, as discussed with the Python magic.

vgvassilev:
We can mark it as complete because @kr-2003 solved it in the clad-feedstock. That’s not a xeus-cpp plugin but a clang plugin which is basically running the kernel with different flags.

anutosh491:
Yeah can be closed I think. Thanks @kr-2003

PR #328, opened 2025-06-03
Final release of JupyterLite-xeus

SylvainCorlay:
After the validation of jupyterlite-0.6 with jupyterlite-xeus, we could release the final for JupyterLite-xeus.


mcbarton (approved):

SylvainCorlay:
JupyterLite-xeus 4.0.0 is exactly the same as 4.0.0rc0.

PR #329, opened 2025-06-03
Add std::cin cells to example notebooks

mcbarton:
# Description

After https://github.com/compiler-research/xeus-cpp/pull/328 std::cin should work for xeus-cpp, so this adds this to the example notebooks.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature


anutosh491:
Something looks off here (probably in both cases)

image

So you've added the cells but not rerun the notebook as such so the configuration stands broken and we can't even see the input box when we do `std::cin`

anutosh491:
Also does it really have to be +900-500 change diff to add 2 cells ? Looks a bit too much :\

anutosh491:
Also not sure what spoiled this 

image


But let's take this opportunity to fix this cell too. We would need to split this into 2 markdown cells so that it is framed correctly

image

SylvainCorlay:
Since `std::cin` interupts the execution of code (preventing a run-all action to complete), maybe we should do it at the very end of the notebook, or just mention it in the markdown cell?

SylvainCorlay (approved):

PR #333, opened 2025-06-04
Fix the case where there is an @ in the absolute path of sysroot and XEUS_CPP_RESOURCE_DIR in Emscripten build

mcbarton:
# Description

I've been working around an issue with the Emscripten build for months locally. If your unlucky and happen to have a @ symbol in your absolute path for SYSROOT_PATH then it will get truncated at this point when preloading, and won't build. This recently got made worse by the introduction of XEUS_CPP_RESOURCE_DIR , which in my case also has an @ in the absolute path. This PR fixes this by creating a symlink in the build directory to these locations, avoiding a @ in the path.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -446,6 +446,19 @@ if(EMSCRIPTEN)
     xeus_cpp_set_kernel_options(xcpp)
     xeus_wasm_compile_options(xcpp)
     xeus_wasm_link_options(xcpp "web,worker")
+    if (SYSROOT_PATH MATCHES "@")
+      execute_process(
+          COMMAND ln -s ${SYSROOT_PATH}
+      )
+      set(SYSROOT_PATH_ORIGINAL ${SYSROOT_PATH})
+      set(SYSROOT_PATH "./sysroot/")
+    endif()
+    if (XEUS_CPP_RESOURCE_DIR MATCHES "@")
+      execute_process(
+          COMMAND ln -s ${XEUS_CPP_RESOURCE_DIR}
+      )
+      set(XEUS_CPP_RESOURCE_DIR "./${CPPINTEROP_LLVM_VERSION_MAJOR}")
+    endif()
     target_link_options(xcpp
         PUBLIC "SHELL: -s USE_SDL=2"
         PUBLIC "SHELL: --preload-file ${SYSROOT_PATH}/include@/include"
@@ -460,6 +473,11 @@ endif()
 
 if(XEUS_CPP_BUILD_TESTS)
     add_subdirectory(test)
+    if (SYSROOT_PATH_ORIGINAL MATCHES "@")
+      execute_process(
+          COMMAND ln -s ${SYSROOT_PATH_ORIGINAL} "${CMAKE_CURRENT_BINARY_DIR}/test/"
+      )
+    endif()
 endif()
 
 # Installation
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -446,16 +446,12 @@ if(EMSCRIPTEN)
     xeus_cpp_set_kernel_options(xcpp)
     xeus_wasm_compile_options(xcpp)
     xeus_wasm_link_options(xcpp "web,worker")
-    if (SYSROOT_PATH MATCHES "@")
-       string(REPLACE "@" "@@" SYSROOT_PATH "${SYSROOT_PATH}")
-    endif()
-    if (XEUS_CPP_RESOURCE_DIR MATCHES "@")
-       string(REPLACE "@" "@@" XEUS_CPP_RESOURCE_DIR "${XEUS_CPP_RESOURCE_DIR}")
-    endif()
+    string(REPLACE "@" "@@" ESCAPED_SYSROOT_PATH "${SYSROOT_PATH}")
+    string(REPLACE "@" "@@" ESCAPED_XEUS_CPP_RESOURCE_DIR "${XEUS_CPP_RESOURCE_DIR}")
     target_link_options(xcpp
         PUBLIC "SHELL: -s USE_SDL=2"
-        PUBLIC "SHELL: --preload-file ${SYSROOT_PATH}/include@/include"
-        PUBLIC "SHELL: --preload-file ${XEUS_CPP_RESOURCE_DIR}@/${CMAKE_INSTALL_LIBDIR}/clang/${CPPINTEROP_LLVM_VERSION_MAJOR}"
+        PUBLIC "SHELL: --preload-file ${ESCAPED_SYSROOT_PATH}/include@/include"
+        PUBLIC "SHELL: --preload-file ${ESCAPED_XEUS_CPP_RESOURCE_DIR}@/${CMAKE_INSTALL_LIBDIR}/clang/${CPPINTEROP_LLVM_VERSION_MAJOR}"
         PUBLIC "SHELL: --preload-file ${XEUS_CPP_DATA_DIR}@/share/xeus-cpp"
         PUBLIC "SHELL: --preload-file ${XEUS_CPP_CONF_DIR}@/etc/xeus-cpp"
         PUBLIC "SHELL: --post-js ${CMAKE_CURRENT_SOURCE_DIR}/wasm_patches/post.js"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -447,17 +447,10 @@ if(EMSCRIPTEN)
     xeus_wasm_compile_options(xcpp)
     xeus_wasm_link_options(xcpp "web,worker")
     if (SYSROOT_PATH MATCHES "@")
-      execute_process(
-          COMMAND ln -s ${SYSROOT_PATH}
-      )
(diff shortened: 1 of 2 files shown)


vgvassilev (commented):
Looks reasonable to me. Let's hear from the other reviewers.

mcbarton:
@anutosh491 @SylvainCorlay can you review this PR for me? All this PR does it create a symlink to the respective paths, in the correct locations it you happen to have an @ in the absolute paths for these directories. Once this symlink has been created, the paths are re adjusted to use a relative path through through these symlinks. Thereby avoiding the issue of path truncation when preloading the directory.

anutosh491:
cc @SylvainCorlay what's your take on this 

Here is my take 

1) This looks like a very special case to me. I don't think its natural to have @ in your path (its unusual and necessarily wrong to be fair)
2) How does one even test this before taking it in ?
3) I searched through emscripten's github and I don't see any issues challenging their design of using the "@" character for preloading. If it was a usual case where people are having issues with having "@" in their path, this would have been raised !

Also I'm not in favour of special handling for every other special character that every other compiler (that we might use) would introduce in future.

I personally think @mcbarton should deal with this locally. I am scared of addressing changes that aren't proven issues fundamentally/universally !

mcbarton:
> cc @SylvainCorlay what's your take on this
> 
> Here is my take
> 
>     1. This looks like a very special case to me. I don't think its natural to have @ in your path (its unusual but not necessarily wrong to be fair)
> 
>     2. How does one even test this before taking it in ? Like I've been tagged for a review and I don't know what to comment on (I can't test this locally)
> 
>     3. I searched through emscripten's github and I don't see any issues challenging their design of using the "@" character for preloading. If it was a usual case where people are having issues with having "@" in their path, this would have been raised !
> 
> 
> Also I'm not in favour of special handling for every other special character that every other compiler (that we might use) would introduce in future.
> 
> I personally think @mcbarton should deal with this locally. I am scared of addressing changes that aren't proven issues fundamentally/universally !

As discussed in our group chat on Discord @anutosh491 (, just because an issue/bug may be non universal, doesn't mean it should be ignored (which seems to be your point 1 and point 3), when a simple solution exists (e.g. creating a symlink, so a relative path can be used). It is not as if we are not using relative paths in some cases already. I know its just about to be removed, but the XEUS_CPP_DATA_DIR uses a relative path https://github.com/compiler-research/xeus-cpp/blob/3a888b62d53b9e289a7688c014fe0a4bc5b69956/CMakeLists.txt#L27 .With regards to you being unable to test the solution locally, I already gave you one in the Discord chat (e.g. rename the micromamba environment names to have an @ in them). With regards to addressing this locally, I already have been doing so for months, and its getting frustrating. I am having to fetch this patch whenever I want to develop a new feature for xeus-cpp, or review a PR. I don't believe I am personally asking for much in this PR, or providing an overly complicated solution.

anutosh491:
> With regards to you being unable to test the solution locally, I already gave you one in the Discord chat (e.g. rename the micromamba environment names to have an @ in them)

Exactly my point !!!

I need to purposefully make something wrong to get this right. Which makes it unnatural. 
When I mean this is a special case, I mean it doesn't appear in the wild and can't be reproduced as it is.

I truly believe if something is in master, it should atleast be a fundamental issue that should be reproducible to some level naturally (stressing on this)

>  already have been doing so for months, and its getting frustrating

I don't see anything wrong with this. I am convinced this is a special case and not a bug.

1) You can always maintain a patch and "git apply" that patch before building
2) You can use "git add -p" and never pick up  and omit that patch while committing.

vgvassilev:
> 1. This looks like a very special case to me. I don't think its natural to have @ in your path (its unusual but not necessarily wrong to be fair)

A university system already uses this. They likely use a valid setup that's not unique to that institution. If we want to position xeus-cpp-lite as a tool for learning we *must* fix such cases, no matter what.

CMakeLists.txt
```diff
@@ -446,6 +446,19 @@ if(EMSCRIPTEN)
...
+      )
+      set(SYSROOT_PATH_ORIGINAL ${SYSROOT_PATH})
+      set(SYSROOT_PATH "./sysroot/")
+    endif()
+    if (XEUS_CPP_RESOURCE_DIR MATCHES "@")
+      execute_process(
+          COMMAND ln -s ${XEUS_CPP_RESOURCE_DIR}
+      )
+      set(XEUS_CPP_RESOURCE_DIR "./${CPPINTEROP_LLVM_VERSION_MAJOR}")
+    endif()
```
anutosh491: Could you paste what does your build/CMakeFiles/xcpp.dir/link.txt look like after this.

CMakeLists.txt
```diff
@@ -446,6 +446,19 @@ if(EMSCRIPTEN)
...
+    if (SYSROOT_PATH MATCHES "@")
+      execute_process(
+          COMMAND ln -s ${SYSROOT_PATH}
+      )
+      set(SYSROOT_PATH_ORIGINAL ${SYSROOT_PATH})
+      set(SYSROOT_PATH "./sysroot/")
+    endif()
+    if (XEUS_CPP_RESOURCE_DIR MATCHES "@")
+      execute_process(
+          COMMAND ln -s ${XEUS_CPP_RESOURCE_DIR}
```
vgvassilev: That won't work on non-unix systems.

SylvainCorlay (approved):

SylvainCorlay:
I am fine with this version. 

From our conversation over chat @mcbarton still needs to test that this resolves the issue on his system.

mcbarton:
I have tested this new solution works on my system this morning. I think this can go in. Will merge sometime later today, unless I hear any objection to this.

CMakeLists.txt
anutosh491: Hey @mcbarton could you quickly paste this for me.

mcbarton: Hi @anutosh491 I'm not sure of its relevance anymore, since I changed to @SylvainCorlay solution since you made this comment, and he labelled it as resolved, but here it is (I have swapped out my actual email, with the word email, as I do not want to and will not post this information publically)

```
/Users/email/micromamba/envs/xeus-cpp-wasm-build/opt/emsdk/upstream/emscripten/em++ -O3 -DNDEBUG -O2 -g0 -s WASM=1 -sWASM_BIGINT -L/lib --bind -Wno-unused-command-line-argument -fexceptions -s MODULARIZE=1 -s EXPORT_NAME=createXeusModule -s EXPORT_ES6=0 -s ASSERTIONS=0 -s ALLOW_MEMORY_GROWTH=1 -s EXIT_RUNTIME=1 -s WASM=1 -s ENVIRONMENT=web,worker -s STACK_SIZE=32mb -s INITIAL_MEMORY=128mb -s WASM_BIGINT -s EXPORTED_RUNTIME_METHODS=[\"FS\",\"ENV\",\"PATH\",\"LDSO\",\"getDylinkMetadata\",\"loadDynamicLibrary\",\"ERRNO_CODES\"] -s FORCE_FILESYSTEM -s MAIN_MODULE=1 -s USE_SDL=2 --preload-file /Users/email/micromamba//envs/xeus-cpp-wasm-build/opt/emsdk/upstream/emscripten/cache/sysroot/include@/include --preload-file /Users/email/micromamba/envs/xeus-cpp-wasm-host/lib/clang/20@/lib/clang/20 --post-js /Users/email/xeus-cpp/wasm_patches/post.js -sWASM_BIGINT @CMakeFiles/xcpp.dir/objects1.rsp -o xcpp.js @CMakeFiles/xcpp.dir/linkLibs.rsp
```

anutosh491: Hmmm I see this 
```
-preload-file /Users/email/micromamba//envs/xeus-cpp-wasm-build/opt/emsdk/upstream/emscripten/cache/sysroot/include@/include
```


So you have a path like `/Users/email/micromamba//envs/` with the double back slash ?

anutosh491: I just want to make sure this isn't messing with the preloading !

mcbarton: That double forward slash is nothing to do with this issue. I have to manually set MAMBA_ROOT_PREFIX which is why this is there (the forward slash automatically gets put there where I auto complete the path), since this environment variable doesn't get set on my Macbook for some reason. 

I tried to fix this MAMBA_ROOT_PREFIX  issue in a PR before, but got told it doesn't matter, because its an environment variable other kernels choose to use, and that it must be something I am doing wrong.

I will now be setting this comment as resolved again.

vgvassilev (approved):
Looks reasonable to me...

anutosh491:
Hi, 

Can you help me with something. Just try out the example notebook before merging this & after merging this !

Obviously we shouldn't see any wrong results but apart from that I'm also concerned about the loading time for the wasm module now that I'm submitting the blog for xeus-cpp-lite.

Hopefully there's no drop in speed. Feel free to test it out and let me know if you see any difference (probably for the smallpt notebook too)

PR #351, opened 2025-06-16
Improve doctest output

mcbarton:
# Description

Currently the output from doctest isn't very useful, as it only gives a number of how many tests pass and fail at the end, and not a list of which tests fail (see image below). It is also hard to disentangle which output to the terminal is from which tests.
image

This PR fixes this by providing a custom doctest config, where we pass specific command line arguments to doctest via the method outlined in doctests documentation (see https://github.com/doctest/doctest/blob/master/doc/markdown/main.md )

In the case of the browser tests that xeus-cpp has, this is the only way to pass these command line arguments. 

Fixes # (issue)

## Type of change

- [x] Bug fix
- [ ] New feature

### DIFF
--- a/test/main.cpp
+++ b/test/main.cpp
@@ -6,5 +6,20 @@
  * The full license is in the file LICENSE, distributed with this software.
  ****************************************************************************/
 
-#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#define DOCTEST_CONFIG_IMPLEMENT
 #include "doctest/doctest.h"
+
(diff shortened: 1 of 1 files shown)


mcbarton:
This is what the improved output from doctest looks like after this PR (see https://github.com/compiler-research/xeus-cpp/actions/runs/15680986691/job/44172145178?pr=351#step:6:7636 for where I took the screenshot)
image

vgvassilev:
The shown screenshot looks much more informative to me.

PR #362, opened 2025-06-30
Enable Windows Emscripten build

mcbarton:
# Description

This PR enable xeus-cpp to be cross compiled to Emscripten on a Windows platform (following Emscripten build instructions in CppInterOp). The 2 changes are as follows

1) Add .bat suffix to embuilder if on Windows (otherwise it uses the one made for Unix systems and errors)
2)  Change \ to / in ESCAPED_XEUS_CPP_RESOURCE_DIR and  ESCAPED_SYSROOT_PATH (\ gets removed from the preloaded path otherwise)

To see these changing allow an Emscripten build of xeus-cpp to work and pass all Emscripten tests, see this PR in CppInterOp https://github.com/compiler-research/CppInterOp/pull/647

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -432,11 +432,8 @@ endif()
 
 if(EMSCRIPTEN)
     # Ensure required SDL2 ports (e.g. SDL2) are built before header preloading
-    if (CMAKE_C_COMPILER MATCHES ".bat")
-      set(embuilder_suffix .bat)
-    endif()
     execute_process(
-        COMMAND embuilder${embuilder_suffix} build sdl2
+        COMMAND embuilder${EMCC_SUFFIX} build sdl2
         RESULT_VARIABLE SDL2_RESULT
     )
     if(NOT SDL2_RESULT EQUAL 0)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -432,8 +432,11 @@ endif()
 
 if(EMSCRIPTEN)
     # Ensure required SDL2 ports (e.g. SDL2) are built before header preloading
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -432,8 +432,11 @@ endif()
 
 if(EMSCRIPTEN)
     # Ensure required SDL2 ports (e.g. SDL2) are built before header preloading
+    if (CMAKE_C_COMPILER MATCHES ".bat")
```
vgvassilev: Is this a better way to detect the if the compiler is running on windows: https://github.com/emscripten-core/emscripten/issues/3191

mcbarton: This is the way we have been detecting we are doing an Emscripten build of Windows within CppInterOp. Its possible we might just be able to use EMCC_SUFFIX from the Emscripten.cmake file, which we use as part of the build. Its probably more reliable too, since it will update with whatever Emscripten decides the suffix should be. I will modify the cmake to use this, and then rerun the workflow in CppInterOp to check it works.

vgvassilev:
Do we want to work in the windows ci here, too?

vgvassilev:
> > Do we want to work in the windows ci here, too?
> 
> @vgvassilev xeus-cpp uses Emscripten forges emsdk, which is only available for very particular platforms. When we asked previously for them to make emsdk available on Windows before the request was rejected. I could add a Windows Emscripten build here too, but we would need to get the upstream emsdk to achieve this. What would you like me to do?

I am not sure what's our path forward. We can merge that PR as it is if it helps with anything. However, it'd probably regress without any way of testing even small parts of this code for that platform..

PR #378, opened 2025-09-25
Drop pin on nlohmann json

mcbarton:
# Description

I was looking through Emscripten forges recipe repo and notice the pin on nlohmann json is being dropped, so this PR does the same for the main repo.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature


anutosh491:
You also need to make the nlohmann_json_abi changes !

anutosh491:
Check the latest recipes on conda-forge and emscripten-forge (and use a git blame if that helps. On phone right now so won't be able to find a link)

mcbarton:
Not sure why nlohmann_json-abi can be found when doing the native build on the Github runners, but not the read the docs build. I will investigate later.

environment-wasm-host.yml
```diff
@@ -1,9 +1,10 @@
 name: xeus-cpp-wasm-host
 channels:
   - https://repo.prefix.dev/emscripten-forge-dev
-  - https://repo.mamba.pm/conda-forge
+  - conda-forge
```
mcbarton: @anutosh491 I have no idea why change this is needed to allow nlohmann_json-abi to be found when creating the environment in the read the docs build, but not the Github runners build, but it works. I just took a punt that was the solution seeing that was what xeus-r has in their environment file https://github.com/jupyter-xeus/xeus-r/blob/main/environment-wasm-host.yml

anutosh491: I made this change few days back 

These are the right channels to use (what Jupyter officially uses on its website for fetching all possible kernels https://jupyter.org/try)

https://github.com/compiler-research/xeus-cpp/pull/373

Let me know if this fixes your issue. if yes, let's move my PR in and next we can rebase your PR keeping only the nlohmann_json change !

anutosh491 (approved):
Okay I'm guessing this solves the build the docs thing. Let's merge it and then you can close my previous Pr I guess ?

environment-wasm-host.yml
JohanMabille: `nlohmann_json-abi` was added recently to conda-forge, which is still hosted on anaconda.com. `https://repo.mamba.pm/conda-forge` is a mirror, and it was probably not updated correctly.

PR #393, opened 2025-10-17
Release 0.8.0

anutosh491:
# Description

This PR finalizes the 0.8.0 patch release of xeus-cpp.

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/include/xeus-cpp/xeus_cpp_config.hpp
+++ b/include/xeus-cpp/xeus_cpp_config.hpp
@@ -11,8 +11,8 @@
 
 // Project version
 #define XEUS_CPP_VERSION_MAJOR 0
-#define XEUS_CPP_VERSION_MINOR 7
-#define XEUS_CPP_VERSION_PATCH 2
+#define XEUS_CPP_VERSION_MINOR 8
+#define XEUS_CPP_VERSION_PATCH 0
 
 // Composing the version string from major, minor and patch
 #define XEUS_CPP_CONCATENATE(A, B) XEUS_CPP_CONCATENATE_IMPL(A, B)


SylvainCorlay:
Semantic versioning works well for scripting languages like Python or JavaScript, but compiled languages must also encode binary compatibility in their version numbers. The system described here communicates both ABI and API compatibility in a single version number.

**Example: Xeus**

The xeus recipe includes:

```yaml
  run_exports:
    - ${{ pin_subpackage('xeus', upper_bound='x.x') }}
``` 

This means "_if you link with `xeus 5.3.1` at build time, your runtime requirement is `xeus >= 5.3.1,<5.4.0`_". This pattern is widely adopted: over 1,500 packages on conda-forge use pin_subpackage. Maintainers often rely on tools like [ABI Laboratory](https://abi-laboratory.pro/) to assess binary compatibility and versioning policies.

**Impact on Downstream Packages**
For a downstream package like xeus-octave:

- Patch updates to Xeus do not require rebuilding xeus-octave.
 - Minor updates to Xeus are not automatically picked up by existing builds, but a new build (with the same recipe) can incorporate the update.
- Major updates to Xeus typically require code changes and a new version of xeus-octave.

Xeus also enforces a strict SOVERSION policy, following the libtool convention.

**Why Strict Minor Version Discipline Matters**

Xeus is a dependency for many packages (xeus-python, xeus-cpp, xeus-octave, xeus-r, xeus-lua, xeus-sql, xeus-robot, xeus-sqlite, xeus-ocaml, xeus-qt, xwidgets, etc.). A minor version bump in Xeus would force a rebuild of all dependent packages. Without this, recent builds of these packages could end up with mismatched minor versions of Xeus, causing conflicts in the solver and preventing co-installation in the same environment.
**Thus, minor version bumps should be used sparingly.**

If we want xeus-cpp to be a common dependency for many packages, I think it is a good idea to follow these conventions by default.

PR #398, opened 2025-10-23
Fixed default mime_bundle_repr

JohanMabille:
Currently, executing `xcpp::display(33)` outputs the adress of the variable instead of its value. This PR fixes this error.

PR #386, opened 2025-09-30
Use check-xeus-cpp target to run C++ tests for both native and Emscripten platform (+ make build instructions consistent across files)

mcbarton:
# Description

Please include a summary of changes, motivation and context for this PR.

This PR makes use of a check-xeus-cpp target to run the C++ tests on both the native and Emscripten platform, and updates the documentation to be consistent across all files (e.g. BUILD_TOOLS_PREFIX vs BUILD_PREFIX)

Fixes # (issue)

## Type of change

Please tick all options which are relevant.

- [ ] Bug fix
- [ ] New feature
- [ ] Added/removed dependencies
- [ ] Required documentation updates

PR #400, opened 2025-10-24
Use node version 22 to run wasm based tests

anutosh491:
# Description

Refer https://github.com/compiler-research/xeus-cpp/pull/398#issuecomment-3441684398

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -74,6 +74,7 @@ jobs:
       - name: Test Emscripten xeus-cpp in browser
         shell: bash -l {0}
         run: |
+          set -e
           cd build/test
           # Fresh install browsers, and run Emscripten tests in them
           # This is to match the Emscripten build instructions, where
@@ -141,7 +142,6 @@ jobs:
       - name: Jupyter Lite integration
         shell: bash -l {0}
         run: |
-          set -e
           micromamba create -n xeus-lite-host jupyterlite-core=0.6 jupyter_server jupyterlite-xeus -c conda-forge
           micromamba activate xeus-lite-host
           jupyter lite build \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -246,6 +246,7 @@ jobs:
       - name: Test Emscripten xeus-cpp in browser
         shell: bash -l {0}
         run: |
+          set -e
           cd build/test
           # Fresh install browsers, and run Emscripten tests in them
           # This is to match the Emscripten build instructions, where
@@ -318,7 +319,6 @@ jobs:
       - name: Jupyter Lite integration
         shell: bash -l {0}
         run: |
-          set -e
           micromamba create -n xeus-lite-host jupyterlite-core=0.6 jupyterlite-xeus -c conda-forge
           micromamba activate xeus-lite-host
           jupyter lite build --XeusAddon.prefix=${{ env.PREFIX }}


.github/workflows/deploy-github-page.yml
```diff
@@ -66,12 +66,15 @@ jobs:
             ..
           emmake make -j ${{ env.ncpus }} install
 
+      - name: Setup latest Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'latest'
```
mcbarton: If the user needs to use the latest node for the Emscripten tests to pass in the referenced PR, then this would needs to be mentioned in the readme. The user may not have the latest node installed, and executing node will get emsdks downloaded version. We should be using the node version which comes with emsdk to be consistent. When we move to a newer emsdk we get a very recent node. Maybe for now we need something in the readme about emsdk install node-24.7.0-64bit

.github/workflows/deploy-github-page.yml
```diff
@@ -140,6 +143,7 @@ jobs:
       - name: Jupyter Lite integration
         shell: bash -l {0}
         run: |
+          micromamba create -n xeus-lite-host jupyterlite-core=0.6 jupyter_server jupyterlite-xeus -c conda-forge
```
mcbarton, line 145: This PR is about using the latest node, so why does it add jupyter_server too? This would make it more consistent with what CppInterOp has, but the goal of the PR doesn't have anything to do with this that I can see.

mcbarton (changes requested):

.github/workflows/deploy-github-page.yml
anutosh491: Check commit 1 error !

test/CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ if(EMSCRIPTEN)
         PUBLIC "SHELL: -s WASM_BIGINT"
         PUBLIC "SHELL: -s ASSERTIONS=0"
         PUBLIC "SHELL: -s ALLOW_MEMORY_GROWTH=1"
+        PUBLIC "SHELL: -s EXIT_RUNTIME=1"
```
mcbarton: The emrun flag we have enables this, and it was removed in this PR https://github.com/compiler-research/xeus-cpp/pull/358 . This is unneeded

anutosh491:
Weird, would you like to take over ?

The goal here was to upgrade the node version cause node 20 gives a memory out of bounds error for #398 , whereas there is nothing wrong with the change/test added there. Upgrading to 23/24 should fix that !

anutosh491:
cc @mcbarton we didn't realize we could simply fetch nodejs24 from conda-forge and address this (makes it a 2 line change and I'll update the readme too)

mcbarton:
@anutosh491 Is node 22 is not new enough? I would prefer the emscripten forge recipes PR goes in now its been approved, and we use the node from the build environment. If you need node 24, then maybe update the emsdk recipe to be node 24 rather than node 22 like I set it. You could fix 2 tasks at the same time if your willing to approve https://github.com/compiler-research/xeus-cpp/pull/386 , since that uses node from tue build environment  (the node being whatever you set it to in the emscripten forge PR), and gives xeus-cpp to have a target which automatically runs the c++ tests in node

mcbarton (changes requested):
We should not be running our tests through a node version added to the xeus-lite-host environment. We should be using the same one which is used during the configure stage of our Emscripten build. Without this your using node 16 through the CMAKE_CROSSCOMPILING_EMULATOR , and then running your tests in node 22. This inconsistency could cause our tests to fail for random reasons which would be hard to debug, and would almost certainly get worse as the node from the build environment gets more and more outdated. What needs to be done is for the node used in the Emscripten forge recipe to be updated to a newer version (to whatever matches the version the upstream emsdk), and a mechanism should be put in place that keeps the MacOS emsdk provided by Emscripten forge upto date with the Linux version.

Although I have no evidence towards this, the emsdk versions not being kept in sync could explain why sometimes in the past we have seen cases where we have had Emscripten jobs fail for one system, but not the other (see https://github.com/compiler-research/xeus-cpp/actions/runs/14780201517 and https://github.com/compiler-research/xeus-cpp/actions/runs/15210235224 for examples where this has happened).

@SylvainCorlay I would like your opinion on this matter.

mcbarton:
> Hey @mcbarton , as Vassil and I discussed on discord (message [link](https://discord.com/channels/@me/1350020222605856869/1435155178587164784)) Node is just a testing framework here. We generally would not care unless it's the same version we get from anywhere (be it emsdk) and unless it contributes somehow to come up with `xcpp.wasm`
> 
> That being said, I've also dropped the osx26 runner for emscripten. We have discussed this in the past on discord. Even if we allow multiple platforms in the CI for emscripten, multiple runners doesn't really contribute here. Our native kernels still use osx15, when we drop those in the future, we can do the same for emscripten and drop osx15 for osx26. But yeah untill then just 1 runner per platform is more than sufficient.
> 
> This PR is now ready to go in on my end.
> 
> Some other small changes in the readme being I've shifted the install step after the build step which is how it should be logically.

Hi @anutosh491 I have read through your latest PR changes and have some questions. Maybe I am missing something, but not sure how changing `emmake make check-xeus-cpp` to `make check-xeus-cpp` makes sense. Don't we need the emmake command always for Emscripten builds, so that it sets the correct environment variables for the Emscripten build? In this case for the test target wasm, js and html are built with Emscripten.

I don't agree that its logical to move the install step to before the browser tests have been run. I don't understand why you would want to install something into your environment that you haven't fully tested. You even put in your changes `Once the test pass, run the install command` in the readme after changing it, forgetting the browser tests are part of the testing. This PR should be kept to its original purpose of updating node to a newer version, and not be used as a catch all for multiple things you see as bugs. This particular change will need to be reverted if you want my approval for the PR. If you insist on this change being part of this PR, then you will need to get @vgvassilev to approve directly.

In the ci section on 'Test Emscripten xeus-cpp in browser' you moved the command which created the `xeus-lite-host` environment. I'm sure what moving where this environment is created has to do with updating node to a newer version, but in the process you have created a bug in the ci. You moved `set -e` from `Test Emscripten xeus-cpp in browser` . Without this `set -e` I am pretty sure that failures in the browser tests will not cause the ci to fail. On the flip side `set -e` will have no effect on the `Jupyter Lite integration` stage of the ci. Please revert this change change with regards to `set -e` at the very least.

On the subject of having just one MacOS job I am not too fussed. I just don't understand why you decided on deleting the more recent MacOS runner. Surely if we are keeping just one then we would want the latest one.

anutosh491:
> In the ci section on 'Test Emscripten xeus-cpp in browser' you moved the command which created the xeus-lite-host environment. I'm sure what moving where this environment is created has to do with updating node to a newer version, but in the process you have created a bug in the ci. You moved set -e from Test Emscripten xeus-cpp in browser . Without this set -e I am pretty sure that failures in the browser tests will not cause the ci to fail. On the flip side set -e will have no effect on the Jupyter Lite integration stage of the ci. Please revert this change change with regards to set -e at the very least.

This is a general improvement. The xeus-lite-host env has nothing to do with the browser tests. I just moved it to where it should be !

anutosh491:
> I think you missed my point that moving set -e along with the creation of the environment creates a bug in the ci. Without the set -e where it was, we won't know when the browser tests fail.

Ahhh okay Sorry, I'll put this back.

anutosh491:
> That target may run that command, but it depends on test_xeus_cpp which is a build target (see

We are just using node as the runtime for the js file here. It does not build anything. I can't explain this any better. You don't need emcmake or emmake untill its a cmake config or a make step !

We were just running `node test_xeus_cpp.js` before this right ? Did we ever need anything from emscripten then ? 
I does not matter !

mcbarton:
> > That target may run that command, but it depends on test_xeus_cpp which is a build target (see
> 
> We are just using node as the runtime for the js file here. It does not build anything. I can't explain this any better. You don't need emcmake or emmake untill its a cmake config or a make step !
> 
> We were just running `node test_xeus_cpp.js` before this right ? Did we ever need anything from emscripten then ? I does not matter !

Even though I still don't understand how its not needed, I'll accept on this point that you are right since the ci passes.

vgvassilev (approved):
LGTM!

PR #404, opened 2025-11-06
Remove unused xeus-cling and xtensor images

mcbarton:
As far as I can tell these images are not being used anywhere in the repo, so this PR removes them.


anutosh491:
> Yes, but why is this different than other libraries?

I think that's because xeus-cling which was maintaining the responsible tagfiles for separate libraries as shown here (https://github.com/jupyter-xeus/xeus-cling/tree/main/etc/xeus-cling/tags.d).  This is the not best approach right ? we don't want to host tagfiles for each library in our code.

Rather how xeus-cpp does it is, if a library author wants to support inline documentation through a notebook, they just need to generate and host tagfiles in their source itself and when we fetch the package the tagfiles land in our prefix from where we can use them.

I think Johan and I discussed this long back but yeah this can be improved on I think for sure ? Some kind of mechanism to generate the required files basically !

JohanMabille (approved):

PR #405, opened 2025-11-06
[ci] Run tests under Valgrind Ubuntu x86

mcbarton:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
           - name: ubu24-analyzers
             os: ubuntu-24.04
             coverage: true
-            extra_cmake_flags: -DCMAKE_BUILD_TYPE=Debug
+            debug: on
             micromamba_shell_init: bash
           - name: ubu22
             os: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


mcbarton:
> Why is this PR needed in the first place ?
> 
> P.S : Asking as someone who is clueless about Valgrind.

The truthfull reason as to why I added this to the ci is to be consistent with all of compiler researches other repos, which run the tests under Valgrind. I cannot list the benefits of Valgrind as I am still in the process of teaching myself how to use it effectively.  I for example cannot tell you if I am suppressing anything that we shouldn't.  @vgvassilev @Vipul-Cariappa or @aaronj0 though may be able to enlighten you though as to why its used in all the other repos, and help you with a review (once they have bandwidth).

I have limited this PR to just running the tests under Valgrind for Ubuntu x86, as running the tests under Valgrind on Ubuntu arm caused an illegal instruction error from Valgrind which doesn't happen locally, so possibly a bug with the runner. I consider this PR now ready for review.

anutosh491:
cc @JohanMabille 
what's your take on this ? 

Asking cause I lack experience with Valgrind :/

mcbarton:
I have now updated the ci to only reference an x86 Valgrind suppression file.

PR #408, opened 2025-11-10
Fix build instructions

mcbarton:
# Description

With all the changes made in the last few weeks, some necessary changes to the build instructions fell through the cracks when getting reviews. This PR fixes them.

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/docs/source/InstallationAndUsage.rst
+++ b/docs/source/InstallationAndUsage.rst
@@ -34,19 +34,17 @@ executing (replace `$CONDA_PREFIX` with a custom installation prefix if need be)
    cd build
    cmake .. -D CMAKE_PREFIX_PATH=$CONDA_PREFIX -D CMAKE_INSTALL_PREFIX=$CONDA_PREFIX -D CMAKE_INSTALL_LIBDIR=lib
    make check-xeus-cpp
+   make install
 
 and
 
 .. code-block:: bash
 
-   cd ../../test
+   cd ../test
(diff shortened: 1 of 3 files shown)


docs/source/InstallationAndUsage.rst
```diff
@@ -34,19 +34,17 @@ executing (replace `$CONDA_PREFIX` with a custom installation prefix if need be)
    cd build
    cmake .. -D CMAKE_PREFIX_PATH=$CONDA_PREFIX -D CMAKE_INSTALL_PREFIX=$CONDA_PREFIX -D CMAKE_INSTALL_LIBDIR=lib
    make check-xeus-cpp
+   make install
```
mcbarton, line 37: I am not moving `make install` here because I think that is where it should be. Without it the kernel tests all error for me, since it assumes the kernels have been installed.

docs/source/InstallationAndUsage.rst
```diff
@@ -34,19 +34,17 @@ executing (replace `$CONDA_PREFIX` with a custom installation prefix if need be)
...
 
 and
 
 .. code-block:: bash
 
-   cd ../../test
```
mcbarton, line 42: This change is needed because we no longer cd into the test folder insider the build folder to run the C++ tests. We stay inside the build folder and run check-xeus-cpp. That means the source test folder is now only one folder up instead of 2.

PR #409, opened 2025-11-11
Add Emscripten compat include paths to wasm kernel specs

anutosh491:
# Description

**Discovered this while moving to emscripten 4-x for xeus-cpp.**

Problem : The following works for our native kernel but not our wasm kernel 
1) native
image
2) wasm
image

## Type of change

- [x] Bug fix
- [ ] New feature


anutosh491:
**What's Happening here ?**

So on native builds, clang automatically discovers <xlocale.h> through the system SDK or libc include paths.
In the Emscripten sysroot, however, legacy headers like <xlocale.h> reside under /include/compat, which isn’t part of the default search list.

How emcc does it 

```
anutosh491@Anutoshs-MacBook-Air xeus-cpp % cat xlocale_test.c
#include <stdio.h>
#include <locale.h>
#include <xlocale.h>

int main(void) {
    locale_t loc = newlocale(LC_NUMERIC_MASK, "C", NULL);

    const char *s = "3.14159";
    long double v = strtold_l(s, NULL, loc);

    printf("Parsed value: %.5Lf\n", v);

    freelocale(loc);
    return 0;
}

anutosh491@Anutoshs-MacBook-Air xeus-cpp % emcc -v xlocale_test.c -o xlocale_test.html
 /Users/anutosh491/work/emsdk/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --sysroot=/Users/anutosh491/work/emsdk/upstream/emscripten/cache/sysroot -DEMSCRIPTEN -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat -v -c xlocale_test.c -o /var/folders/m1/cdn74f917994jd99d_2cpf440000gn/T/emscripten_temp_v4f2uvqa/xlocale_test.o
clang version 22.0.0git (https:/github.com/llvm/llvm-project c13ac9cadf1f9b4fa886b82d1e84a5feb0439023)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/anutosh491/work/emsdk/upstream/bin
 (in-process)
 "/Users/anutosh491/work/emsdk/upstream/bin/clang-22" -cc1 -triple wasm32-unknown-emscripten -emit-obj -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name xlocale_test.c -mrelocation-model static -mframe-pointer=none -ffp-contract=on -fno-rounding-math -mconstructor-aliases -target-cpu generic -fvisibility=hidden -debugger-tuning=gdb -fdebug-compilation-dir=/Users/anutosh491/work/xeus-cpp -v -fcoverage-compilation-dir=/Users/anutosh491/work/xeus-cpp -resource-dir /Users/anutosh491/work/emsdk/upstream/lib/clang/22 -D EMSCRIPTEN -isysroot /Users/anutosh491/work/emsdk/upstream/emscripten/cache/sysroot -internal-isystem /Users/anutosh491/work/emsdk/upstream/lib/clang/22/include -internal-isystem /Users/anutosh491/work/emsdk/upstream/emscripten/cache/sysroot/include/wasm32-emscripten -internal-isystem /Users/anutosh491/work/emsdk/upstream/emscripten/cache/sysroot/include -ferror-limit 19 -fmessage-length=141 -fgnuc-version=4.2.1 -fskip-odr-check-in-gmf -fignore-exceptions -fcolor-diagnostics -iwithsysroot/include/fakesdl -iwithsysroot/include/compat -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr -o /var/folders/m1/cdn74f917994jd99d_2cpf440000gn/T/emscripten_temp_v4f2uvqa/xlocale_test.o -x c xlocale_test.c
clang -cc1 version 22.0.0git based upon LLVM 22.0.0git default target 
ignoring nonexistent directory "/Users/anutosh491/work/emsdk/upstream/emscripten/cache/sysroot/include/wasm32-emscripten"
 /Users/anutosh491/work/emsdk/upstream/emscripten/cache/sysroot/include/compat
 /Users/anutosh491/work/emsdk/upstream/lib/clang/22/include
 /Users/anutosh491/work/emsdk/upstream/emscripten/cache/sysroot/include
End of search list.
```

This shows 3 paths 
i) `sysroot/include` and `lib/clang/22/include` (resource dir) which we address as of now
ii) What we don't address is `sysroot/include/compat` !!!!


**As can be seen this is addressed through** `-iwithsysroot/include/compat`

This PR adds the same (check right side of the img below as now we're able to search through `compat` for `xlocale.h`)

image

test/test_interpreter.cpp
```diff
@@ -1136,3 +1136,36 @@ TEST_CASE("C and C++ stdout/stderr capture")
     REQUIRE(captured_err.find("C stderr") != std::string::npos);
     REQUIRE(captured_err.find("C++ stderr") != std::string::npos);
 }
+
+#if defined(XEUS_CPP_EMSCRIPTEN_WASM_BUILD)
+TEST_CASE("headers found in sysroot/include/compat")
+{
```
anutosh491: Added a simple test confirming we can access headers from the location of interest. Technically only need for the wasm build cause its straightforward for the native kernel (so probably not needed)

But the cc1 command as shown in our tests now cover all 3 paths that emcc covers as I've pointed above 
```
clang version 20.1.8 (https://github.com/emscripten-forge/recipes 26347a2820b7a44b0329d3b9bd524c56ff3cdce7)
Failed to detect the resource-dir
exec: : Function not implemented
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: 
 "" -cc1 -triple wasm32-unknown-emscripten -emit-obj -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name "<<< inputs >>>" -mrelocation-model static -mframe-pointer=none -ffp-contract=on -fno-rounding-math -mconstructor-aliases -target-cpu generic -debugger-tuning=gdb -fdebug-compilation-dir=/ -v -fcoverage-compilation-dir=/ -resource-dir /home/runner/micromamba/envs/xeus-cpp-wasm-host -internal-isystem /include/wasm32-emscripten/c++/v1 -internal-isystem /include/c++/v1 -internal-isystem /home/runner/micromamba/envs/xeus-cpp-wasm-host/include -internal-isystem /include/wasm32-emscripten -internal-isystem /include -std=c++14 -fdeprecated-macro -ferror-limit 19 -fvisibility=default -fgnuc-version=4.2.1 -fskip-odr-check-in-gmf -fcxx-exceptions -fexceptions -iwithsysroot/include/compat -fincremental-extensions -o "<<< inputs >>>.o" -x c++ "<<< inputs >>>"
clang -cc1 version 20.1.8 based upon LLVM 20.1.8 default target wasm32-unknown-emscripten
ignoring nonexistent directory "/include/wasm32-emscripten/c++/v1"
ignoring nonexistent directory "/home/runner/micromamba/envs/xeus-cpp-wasm-host/include"
ignoring nonexistent directory "/include/wasm32-emscripten"
 /include/compat
 /include/c++/v1
 /include
End of search list.
```

src/xinterpreter.cpp
```diff
@@ -45,6 +45,12 @@ void* createInterpreter(const Args &ExtraArgs = {}) {
     ClangArgs.push_back("-isystem");
     ClangArgs.push_back(CxxInclude.c_str());
   }
+
+#ifdef __EMSCRIPTEN__
+  ClangArgs.push_back("-Xclang");
+  ClangArgs.push_back("-iwithsysroot/include/compat");
```
vgvassilev: Doesn't that need to go to the emscripten toolchain file in clang ideally?

anutosh491: Hmm not sure (possibly no ?)
 
Because the `compat/` headers are added by `emcc` via `-Xclang -iwithsysroot/include/compat` only at the tool level at the very end https://github.com/emscripten-core/emscripten/blob/3622274db64d8ce706690197d1c24cd02715d937/tools/compile.py#L144

The upstream Clang WebAssembly toolchain ( [Webassembly.cpp](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/WebAssembly.cpp) looks target-generic atleast for the path based utilties (for eg [AddClangSystemIncludeArgs](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/WebAssembly.cpp#L473), [addLibCxxIncludePaths](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/WebAssembly.cpp#L563C19-L563C40)) and won’t encode Emscripten SDK layout quirks like `include/compat` just yet

But I do see emscripten specific exception handling (for [eg](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/WebAssembly.cpp#L415-L417)) so maybe we can have emscripten specific stuff here ?!

anutosh491: Just thinking a bit more I think no 

1) The emscripten specifc exception handling (-enable-emscripten-*) flags I mentioned are present so the compiler backend can reproduce Emscripten’s codegen semantics.
2) The file does not have a reason to know that “Emscripten’s sysroot layout has an /include/compat folder.” So Sysroot layout quirks like /include/compat belong in the Emscripten wrapper (emcc) or, for direct Clang users, must be added manually.

Something like 
```
# Path to the Emscripten sysroot from the SDK
SYSROOT=$HOME/work/emsdk/upstream/emscripten/cache/sysroot

/path/to/your/clang \
  --target=wasm32-unknown-emscripten \
  -isysroot $SYSROOT \
  -Xclang -iwithsysroot/include/compat \
  -c a.cpp -o a.o
  ```

vgvassilev: If that include path is something that is required to compile emscripten code this should be part of the clang emscripten toolchain file.

anutosh491: Well yeah that's what I was pointing out that technically in the toolchains folder we don't see an "emscripten" based file but more of a "webassembly" target-generic file 

https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/WebAssembly.cpp

So not sure this can handle emscripten specifics but let me ask around. Sam Clegg from emscripten should be able to help me out here.

anutosh491: Hey @vgvassilev,

As mentioned by Sam Clegg here, 

https://github.com/llvm/llvm-project/issues/167669#issuecomment-3522944512

He does agree that it would be nice if the clang toolchain can cover most of it but as of now we would have to replicate the emcc driver logic ourselves.

And as in case of xeus-cpp-lite, we can technically say that the driver is the `argv` array in `wasm_kernel.json` .... so I guess its best to handle this information not at the interpreter level in `CreateInterpreter` through an `#ifdef Emscripten` rather a level above at the kernel.

vgvassilev: Okay, if this is not needed/wanted in clang then we can move forward as a kernel option as you suggested.

test/test_interpreter.cpp
mcbarton: If its trivially going to pass for the native case, then just to reduce the diff there is no reason to have the 
`#if defined(XEUS_CPP_EMSCRIPTEN_WASM_BUILD)`

anutosh491: This is an emscripten specific tests cause `sysroot/include/compat` is specific to an emscripten build. Check the argv vector ! 

When I meant we don't need it for the native case, I meant the native case would need a separate test and that would be pretty trivial (we won't check including every other header right ?)

PR #411, opened 2025-11-12
Revert "Fix missing C stderr output (fprintf(stderr, ...) not shown) …(#402)"

mcbarton:
# Description

This reverts commit 5860e959a9fb6ea170e124e5e16e0ed05bd95117. Although with this commit xeus-cpps Emscripten ci passes, it made CppInterOps Emscripten ci fail. The Emscripten ci in CppInterOp has failed ever since this commit went in. I initially discounted it as the source of the problem, despite it being the obvious difference between when the ci passed and when it failed, but that is because I made the silly mistake of forgetting there is a big difference between `git reset --soft HEAD~` and `git reset --hard HEAD~` when undoing the last commit of a repo. I opened up a dummy repo in CppInterOp here https://github.com/compiler-research/CppInterOp/pull/749/files which hard resets xeus-cpps latest commit to show the Emscripten ci passes again without it. See here for the workflow run of that PR https://github.com/compiler-research/CppInterOp/actions/runs/19310693489 .


mcbarton:
Since this PR is a direct revert of a commit to get CppInterOps Emscripten ci passing again I am inclined to ignore the git clang format and git clang review workflows.

mcbarton:
> Since this PR is a revert of a commit to get CppInterOps Emscripten ci passing again I am inclined to ignore the git clang format and git clang review workflows.

I have clicked resolved on all git clang review comments due to the reason mentioned here.

anutosh491 (changes requested):
Have we found the fundamental reason as to how this results into a failure ? 

I don't think I'm comfortable with a revert without knowing what exactly happens.

This was not an emscripten build specific PR right ? 
It was solving a bug in both the native and the wasm kernel. And cppinterop's CI for the native case passes correct or does it not ? 

Cause if it does, then there is no difference in this aspect between the latest build of cppinterop from a package manager and the latest main that can cause this right ? Which means the bug fix might not  "technically" be the reason for the failure. 

On top of it, the PR didn't mess with any build flags/configs, so this raises more questions.

anutosh491:
Can you isolate the error locally ? Cause I haven't been able to 

I built latest cppinterop using emsdk 3.1.73 (not the patched one) and made use of it to build `test_xeus_cpp.js` and I don't see an error using it locally. Everything passes as expected.

Not sure why it fails for you if it does, but let's experiement around I guess 
Comment out all tests except the one you think is buggy and re-run test_xeus_cpp.js against node 22. And tell me what do you see ?

anutosh491:
Hey @vgvassilev 

These are the reasons why I am not comfortable with this 

1) @mcbarton claims he has tried it locally. But I don't see the result being pasted anywhere.I tried the same last week and this morning too and my attempt gives me the exact same result as I had pasted this [here](https://github.com/compiler-research/xeus-cpp/pull/411#issuecomment-3525614512)

My attempt uses 

- unpatches emsdk 3.1.73 
- building llvm 20.1.8, building latest cppinterop and building latest xeus-cpp

2) The fix was a platform independent generic fix for both native and wasm kernels. 

- No flag/ build-config etc was touched for any platform
- if the test fails for the wasm kernel it has to fail for the native kernel too (I even asked if this is what's happening above, but got no reply back)

Hence answering to your question

> We know that in happens in this commit, right?

I don't think we can say that just yet.

3) Other points being

- I'm not sure the LLVM logic applies here. Being a monorepo that doesn't run all tests in the CI, it's fair to revert anything that's alarming. But in this case we have 2 separate repos. 
- I know this feature is being put to use through xeus-cpp's deployment link, so untill we have concrete proof, reverting this seems unfair and just looks like the easy way out.

Let's figure this out together here : https://github.com/compiler-research/CppInterOp/pull/749

anutosh491:
Made a comment for the same : https://github.com/compiler-research/CppInterOp/pull/749#issuecomment-3555841500

Issue #259, opened 2025-02-25
Frame tests for xeus-cpp-lite's example notebook


anutosh491:
We need a way to test this change. Can you figure out how to write a test that runs as part of our ci?

_Originally posted by @vgvassilev in https://github.com/compiler-research/xeus-cpp/pull/258#pullrequestreview-2640055171_

anutosh491:
Hi, 

So we have an [example notebook](https://compiler-research.org/xeus-cpp/lab/index.html) that it a demonstration of everything that runs through xeus-cpp-lite (https://compiler-research.org/xeus-cpp/lab/index.html)

We need a way to test the same through our CI.

anutosh491:
Viable options 

1) One possibility could be to use galata and playwright, just like JupyterLab does for its UI tests (very cumbersome to maintain.... basically, it takes scrennshots of your execution and compares resulting images)

2) We can save an executed notebook as the reference one. Then execute the same notebook in the CI (still with playwright), "save it", and the compare the notebook json files

3) Take inspiration from what was done to test out the example notebook for the native case of xeus-cpp. And try to extrapolate this to xeus-cpp-lite. (https://github.com/compiler-research/xeus-cpp/pull/142/files)

cc @tharun571 would you be interested to work on this ? As might end up having quite some relevance to what you did !

anutosh491:
I'll leave the option 2/3 for Tharun to explore as there might be some overlap.

But for starters what we can do is maybe replicate the UI tests that jupyterlite-xeus is running (currently for xeus-python)

1) This line tests the xeus-python kernel with a print('ok') 
https://github.com/jupyterlite/xeus/blob/main/ui-tests/tests/jupyterlite_xeus.spec.ts#L34

2) Then a screenshot is taken to compare the result 
https://github.com/jupyterlite/xeus/blob/main/ui-tests/tests/jupyterlite_xeus.spec.ts-snapshots/jupyter-xeus-execute-default-linux.png

we can just copy the whole ui-tests folder and use it where we want, replacing the xeus-python test with what we want.

cc @kr-2003 

Would you like experimenting with this line of approach ! For now we should be able to get a simple `std::cout << "Hello World" << std::endl` working

vgvassilev:
We have the approach of CppInterOp where we still wait for your input.

anutosh491:
> We have the approach of CppInterOp where we still wait for your input.

[...]

PR #415, opened 2025-11-19
Run notebooks in Jupyter Lite as part of ci (Emscripten xeus-cpp automated testing)

mcbarton:
# Description

This PR is the first in a series of PRs to add automated testing to check changes don't break xeus-cpps existing jupyter lite notebooks, in a ci friendly way. The workflow for checking nothing is broken

First run python script with driver option set to browser you want to run the notebook in (the ci will cover all options, firefox, chome and safari). This should be run while the jupyter server command is running, with --settings-overrides=overrides.json as the first option to provide the download button.

This will download the notebook once it has finished running (the reason for adding overrides.json to add Download option to notebook menu bar). This config to add this was taken from https://jupyterlite.readthedocs.io/en/latest/howto/configure/settings.html . Right clicking the file, finding the download option and clicking didn't work consistently, and never after the notebook had finished running in the case of Safari.

Final step is compare the fully executed notebook in repo (which will be used as referenced) against the fully run notebook downloaded as part of the python script using nbdiff which comes from jupyters nbdime https://github.com/jupyter/nbdime .

If the notebooks are the same nbdiff outputs nothing and the ci carries on. If they are different, nbdiff returns a non zero value and the ci stops with the outputted diff. 

This idea is effectively option 2 suggested here https://github.com/compiler-research/xeus-cpp/issues/259#issuecomment-2681285987, but I am using Selenium instead of Playwright. I have checked the above steps locally, but am setting this PR to draft until I have had a chance to update the documentation and ci to do these automated notebook tests.

This PR will add the 

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [x] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -137,21 +137,22 @@ jobs:
           fi
         timeout-minutes: 4
 
-      - name: Jupyter Lite integration
-        shell: bash -l {0}
-        run: |
-          micromamba create -n xeus-lite-host jupyter_server jupyterlite-xeus -c conda-forge
-          micromamba activate xeus-lite-host
-          jupyter lite build \
-              --XeusAddon.prefix=${{ env.PREFIX }} \
-              --XeusAddon.mounts="${{ env.PREFIX }}/share/xeus-cpp/tagfiles:/share/xeus-cpp/tagfiles" \
-              --XeusAddon.mounts="${{ env.PREFIX }}/etc/xeus-cpp/tags.d:/etc/xeus-cpp/tags.d" \
-              --contents README.md \
-              --contents notebooks/xeus-cpp-lite-demo.ipynb \
-              --contents notebooks/tinyraytracer.ipynb \
-              --contents notebooks/images/marie.png \
-              --contents notebooks/audio/audio.wav \
-              --output-dir dist
+      - name: Serve Jupyter Lite website
+        uses: ./.github/actions/Jupyter-serve
+
+      - name: Test C++23 kernel in xeus-cpp-lite-demo.ipynb
+        uses: ./.github/actions/Emscripten-Notebook-Tests
+        with:
+          notebook: "xeus-cpp-lite-demo.ipynb"
+          kernel: "C++23"
+        timeout-minutes: 5
+
+      - name: Test C++23 kernel in tinyraytracer.ipynb
+        uses: ./.github/actions/Emscripten-Notebook-Tests
+        with:
+          notebook: "tinyraytracer.ipynb"
+          kernel: "C++23"
+        timeout-minutes: 5
 
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v4
--- a/scripts/automated-notebook-run-script.py
+++ b/scripts/automated-notebook-run-script.py
@@ -1,0 +1,330 @@
+import argparse
+from selenium import webdriver
+from selenium.webdriver.chrome.options import Options as ChromeOptions
+from selenium.webdriver.firefox.options import Options as FirefoxOptions
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.by import By
+from selenium.webdriver.common.keys import Keys
+import time
+import sys
+
+
+def cell_is_waiting_for_input(driver):
+    """
+    This function returns true if Jupyter is currently waiting the user to enter
+    text in a box.
+    """
+    try:
+        return driver.find_element(By.CSS_SELECTOR, ".jp-Stdin-input").is_displayed()
+    except Exception:
+        pass
+
+    return False
+
+
+def wait_for_idle_status(driver, current_cell, start_time, timeout):
+    """
+    This function checks whether the kernel is Idle. Used to decide when to move
+    onto executing the next cell. The 0.01 seconds sleep between checks, is to limit
+    the number of checks per second. After the kernel has gone Idle, we print the contents
+    of the cell (source and output) to the terminal.
+    """
+    while (
+        "Idle"
+        not in driver.find_elements(By.CSS_SELECTOR, "span.jp-StatusBar-TextItem")[
+            2
+        ].text
+    ):
+        elapsed = time.time() - start_time
+        # This timeout is provided in case the notebppks stalls during
+        # its execution.
+        if elapsed > timeout:
+            print(
+                f"Timeout reached ({elapsed:.1f} seconds). Stopping Notebook execution."
+            )
+            sys.exit(1)
+        time.sleep(0.01)
+
+    print(current_cell.text)
+
+def rewind_to_first_cell(cell):
+    while True:
+        prev = cell.find_elements(
+            By.XPATH,
+            "preceding-sibling::div[contains(@class,'jp-Notebook-cell')][1]"
+        )
+        if not prev:
+            return cell
+        cell = prev[0]
+
+    
+
+def run_notebook(driver, notebook_area, args):
+    """This functions runs all the cells of the notebook"""
+    print("Running Cells")
+    start_time = time.time()
+    current_cell = driver.find_element(
+        By.CSS_SELECTOR, ".jp-Notebook-cell.jp-mod-selected"
+    )
+
+    current_cell = rewind_to_first_cell(current_cell)
+    
+    while True:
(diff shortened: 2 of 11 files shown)


mcbarton:
See here https://github.com/compiler-research/xeus-cpp/actions/runs/19553412423/job/55990268776?pr=415#step:7:605 for an example of what the output looks like when the notebook it gets once finished running being different to the reference notebook (in this case the difference comes from the fact the notebook didn't have enough time to finish running on the Github runner)

Here is the output for when the workflow output cannot be seen anymore


```
--- notebooks/smallpt.ipynb  2025-11-20 22:31:08.104731
+++ /home/runner/Downloads/smallpt.ipynb  2025-11-20 22:37:33.795913
## replaced (type changed from int to NoneType) /cells/5/execution_count:
-  5
+  None
## modified /cells/5/outputs/0/text:
@@ -3,6 +3,3 @@
 [DEBUG] Dimensions: 320x240, Samples: 16
 [DEBUG] SDL surface created successfully
 [DEBUG] The image should be on your screen soon
-[DEBUG] Mapped pixel buffer to surface
-[DEBUG] Surface saved to render.bmp
-[DEBUG] Surface created successfully
## deleted /cells/5/outputs/1-2:
-  output:
-    output_type: display_data
-    data:
-      image/bmp: Qk16sAQA...<snip base64, md5=d8c4849298c46ca8...>
-  output:
-    output_type: stream
-    name: stdout
-    text:
-      [DEBUG] Surface freed
-      [DEBUG] Exiting main
```

scripts/automated-notebook-run-script.py
```diff
@@ -0,0 +1,199 @@
...
+    if args.driver == "chrome":
+        options = ChromeOptions()
+        options.add_argument("--headless")
+        options.add_argument("--no-sandbox")
+        driver = webdriver.Chrome(options=options)
+
+    elif args.driver == "firefox":
+        options = FirefoxOptions()
+        options.add_argument("--headless")
+        driver = webdriver.Firefox(options=options)
```
mcbarton, line 286: Note to self, these headless options should be enabled by command line argument, in case the user locally wants to see the notebook running in a gui. Only needed due to ci needing them run in a headless browser

scripts/automated-notebook-run-script.py
```diff
@@ -0,0 +1,200 @@
...
+        actions.move_to_element(run_menu).pause(0.1).click().perform()
+        time.sleep(0.5)
+        print("Click on Run All Cells")
+        run_all_menu = wait.until(
+            EC.visibility_of_element_located(
+                (By.XPATH, "//li[normalize-space()='Run All Cells']")
+            )
+        )
+        actions.move_to_element(run_all_menu).click().perform()
+        time.sleep(200)
```
mcbarton: These time.sleeps to run all the cells will depend on the power of the cpu running the tests. Adjusted for now to be long enough that the notebook has enough time to run on the Github runners. Locally these times can be a lot shorter. 

Long term this should be switched out without something which determines if the kernel has gone idle, and aborts if he has taken too long (assumed notebook has stalled at this point). This I feel can be done as part of subsequent PR though.

mcbarton:
> Hi,
> 
> I would recommend starting out with testing the demo notebook instead of the smallpt notebook.
> 
> Testing the smallpt notebook doesn't really add value. Our smallpt notebook gives an impression that SDL is usable in JupyterLite but it's just a gimmick to be fair 😅 .
> 
> if you check the notebook, we use SDL only to render to a static bitmap instead of using it for any "dynamic" opengl canvas related stuff. As xeus-cpp is run in the worker thread and we need the main thread for any events or dynamic stuff .... we're just masking SDL to convert a raw stream to a bitmap (we don't need sdl for this)
> 
> **So testing the smallpt notebook != SDL works through xeus-cpp-lite. No events, no dynamics, no canvases**
> 
> For events (mouse/keyboard etc), dynamic canvases ... we probably can put the jupyter widgets (ipywidgets, xwidgets) to use for now.
> 
> Check example notebooks here for the above. https://derthorsten.github.io/pyb2d3/lite/lab/index.html

Hi @anutosh491 

For I am just trying to show a technique for running any notebook in Jupyter lite using Selenium. I am not too fussed at the moment that it isn't actually making use of SDL. I picked smallpt just because its the easiest case to start off with since it requires no user input besides executing all the cells. The ci now shows it works for Firefox and Chrome. It works locally for Safari too. I am just debugging what is going on with the download part for this browser, since it doesn't appear to be downloading the notebook for this browser in the ci. Once I have this working completely in the ci, my next goal is to work on the normal demo notebook, and add to the script so it can deal with the case when a notebook requests user input (std::cin case). The next stage I was hoping we could make a series of reference notebooks to run through the ci which check we don't break things between PRs (these don't necessarily need to be ones we deploy, and can be in a separate compiler research repo). Basically the script will be modified for increasingly difficult to execute notebooks until we have something that can deal with most notebooks we give it.

scripts/enable-downloads-safari-github-ci.py
```diff
@@ -0,0 +1,33 @@
+import pyautogui
```
mcbarton: The reason I couldn't download the notebook after running the notebook in Safari is because of the following pop up
image
You cannot click allow, since it is running in Safari driver. I tried both locally and in the ci.

The solution is this python script, which makes is so Safari doesn't ask if you want to allow Downloads from any site (outside of Safari driver), e.g. the end result is this page at the end of the script
image

mcbarton:
Benchmark table to be filled in once the ci is passing all jobs consistently (hoping to be soon)

Regardless of which C++kernel, browser or platform run on the xeus-cpp-lite-demo notebook finishes almost as soon as all the cells have been executed

Results for smallpt notebook (time in ci to finish executing notebook after last cell run)

| Kernel             | Platform        | Safari             |  Chrome         |  Firefox           | 
| -------------  | ------------- | ------------- | ------------- | -------------  |
| C++17             | MacOS arm   |                       |                        |                        |
| C++20            | MacOS arm   |                        |                       |                        |
| C++23            | MacOS arm    |                       |                       |                        |
| C++17             | Ubuntu X86   |                       |                       |                        |
| C++20            | Ubuntu X86   |                       |                       |                        |
| C++23            | Ubuntu X86   |                       |                       |                        |

scripts/automated-notebook-run-script.py
```diff
@@ -0,0 +1,308 @@
...
+        cancelable: true,
+        composed: true,   // IMPORTANT for shadow DOM
+        view: window
+        }));
+        });
+        """,
+        download_button,
+    )
+
+    time.sleep(1)
```
mcbarton, line 196: This is almost certainly not to best way to find the download button and click it, but this is the only reliable solution I could find.

mcbarton:
Once we have a example c notebook merged in, I can expand this work in a future PR to add testing of the C kernel. Also as the url for Jupyter Lab is http://127.0.0.1:8888 (same as Jupyter Lite but a different port), I can expand this work in the future so we can have Jupyter Lab testing as part of the ci (and native kernel testing even without the CppInterOp C/C++ interface)

anutosh491:
Hi @mcbarton , 

I am guessing the PR is ready on your end ? I have written to a couple JupyterLite maintainers who shall hopefully help us with the reviews here soon !

vgvassilev (commented):
Is that the simplest, most minimal ci setup that we can have?

.github/workflows/deploy-github-page.yml
```diff
@@ -137,21 +137,50 @@ jobs:
...
+          notebook: "tinyraytracer.ipynb"
+          kernel: "C++20"
+        timeout-minutes: 5
+
+      - name: Test C++23 kernel in tinyraytracer.ipynb
+        uses: ./.github/actions/Emscripten-Notebook-Tests
+        with:
+          notebook: "tinyraytracer.ipynb"
+          kernel: "C++23"
+        timeout-minutes: 5
```
mcbarton, line 183: @vgvassilev The only way I can think to reduce the ci in this PR is to reduce this part so that it only runs the `xeus-cpp-lite-demo.ipynb` and `tinyraytracer.ipynb` kernels with the C++23 kernel (this change in main.yml too). Is that what you had in mind?

mcbarton: @vgvassilev pinging for review. I've tried to respond to your previous comment here https://github.com/compiler-research/xeus-cpp/pull/415#pullrequestreview-4031318969 , with my response here https://github.com/compiler-research/xeus-cpp/pull/415#issuecomment-4160905379 and my comment above.

Issue #413, opened 2025-11-17
Undefined symbols when building xcpp.js for emsdk 4.0.9 build


mcbarton:
The following undefined symbol errors occur when building xcpp.js using emsdk 4.0.9 . Found in CppInterOp emsdk 4.0.9 PR here https://github.com/compiler-research/CppInterOp/actions/runs/19443476160/job/55632348748?pr=753#step:14:161

```
error: undefined symbol: _embind_register_bindings (referenced by root reference (e.g. compiled C/C++ code))
warning: To disable errors for undefined symbols use `-sERROR_ON_UNDEFINED_SYMBOLS=0`
warning: __embind_register_bindings may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _embind_register_class (referenced by root reference (e.g. compiled C/C++ code))
warning: __embind_register_class may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _embind_register_class_constructor (referenced by root reference (e.g. compiled C/C++ code))
warning: __embind_register_class_constructor may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _embind_register_class_function (referenced by root reference (e.g. compiled C/C++ code))
warning: __embind_register_class_function may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _embind_register_function (referenced by root reference (e.g. compiled C/C++ code))
warning: __embind_register_function may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _emval_as (referenced by root reference (e.g. compiled C/C++ code))
warning: __emval_as may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _emval_call (referenced by root reference (e.g. compiled C/C++ code))
warning: __emval_call may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _emval_call_method (referenced by root reference (e.g. compiled C/C++ code))
warning: __emval_call_method may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _emval_decref (referenced by root reference (e.g. compiled C/C++ code))
warning: __emval_decref may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
[...]

PR #416, opened 2025-11-20
Update wasm exception handling model (migration to emsdk 4-x)

anutosh491:
# Description

Hi, 

Today while building xeus-cpp on emscripten-forge's emscripten-4x branch, I saw this 

image

This was expected because 

1) with emsdk 3.1.73 we are using the Emscripten EH model (which is pulls in the invoke_*** functions/symbols)
2) with emsdk 4-x, emscripten promotes using the wasm EH model. Hence emscripten-forge's emscripten-4x branch builds all hosted packages (xeus, llvm and everything involved using `-fwasm-exceptions` and `-sSUPPORT_LONGJMP=wasm`)
3) We need to replicate the same using clang's webassembly toolchain (https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/WebAssembly.cpp#L357-L452)

- using `-fwasm-exceptions` pulls in the following 
```
-target-feature +exception-handling
-target-feature +multivalue
-target-feature +reference-types
-exception-model=wasm
-mllvm -wasm-enable-eh
```
- and `-sSUPPORT_LONGJMP=wasm` which is an emcc specific flag boils down to `-mllvm -wasm-enable-sjlj` which can be used through clang.

So basically we end up replicating the same result but with a different exception handling model.
image

## Type of change

- [ ] Bug fix
- [ ] New feature

### DIFF
--- a/.github/workflows/deploy-github-page.yml
+++ b/.github/workflows/deploy-github-page.yml
@@ -45,7 +45,10 @@ jobs:
       - name: Build and test xeus-cpp in node, then install
         shell: bash -l {0}
         run: |
-          micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32
+          micromamba create -f environment-wasm-host.yml \
+          --platform=emscripten-wasm32 \
+          -c https://prefix.dev/emscripten-forge-4x \
+          -c https://prefix.dev/conda-forge
 
           mkdir build
           pushd build
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -230,7 +230,10 @@ jobs:
       - name: Build and test xeus-cpp in node, then install
         shell: bash -l {0}
         run: |
-          micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32
+          micromamba create -f environment-wasm-host.yml \
+          --platform=emscripten-wasm32 \
+          -c https://prefix.dev/emscripten-forge-4x \
+          -c https://prefix.dev/conda-forge
           
           mkdir build
           pushd build
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,11 @@ micromamba activate xeus-cpp-wasm-build
 You are now in a position to build the xeus-cpp kernel. You build and test it in node by executing the following. Once the test pass, run the install command.
 
 ```bash
-micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32
+micromamba create -f environment-wasm-host.yml \
+--platform=emscripten-wasm32 \
+-c https://prefix.dev/emscripten-forge-4x \
+-c https://prefix.dev/conda-forge
+
 mkdir build
 cd build
 export BUILD_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
--- a/README.md
+++ b/README.md
@@ -89,7 +89,11 @@ micromamba activate xeus-cpp-wasm-build
 
 You are now in a position to build the xeus-cpp kernel. You build and test it in node by executing the following. Once the test pass, run the install command.
 ```bash
-micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32
+micromamba create -f environment-wasm-host.yml \
+--platform=emscripten-wasm32 \
+-c https://prefix.dev/emscripten-forge-4x \
+-c https://prefix.dev/conda-forge
+
 mkdir build
 cd build
 export BUILD_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
--- a/docs/source/InstallationAndUsage.rst
+++ b/docs/source/InstallationAndUsage.rst
@@ -70,7 +70,11 @@ You are now in a position to build the xeus-cpp kernel. You build and test it in
 
 .. code-block:: bash
 
-    micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32
+    micromamba create -f environment-wasm-host.yml \
+    --platform=emscripten-wasm32 \
+    -c https://prefix.dev/emscripten-forge-4x \
+    -c https://prefix.dev/conda-forge
+
     mkdir build
     cd build
     export BUILD_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-cpp-wasm-build
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -19,7 +19,10 @@ eval "$__mamba_setup";
 alias micromamba="$MAMBA_EXE";
 micromamba create -f environment-wasm-build.yml -y;
 micromamba activate xeus-cpp-wasm-build;
-micromamba create -f environment-wasm-host.yml --platform=emscripten-wasm32;
+micromamba create -f environment-wasm-host.yml \
+--platform=emscripten-wasm32 \
+-c https://prefix.dev/emscripten-forge-4x \
+-c https://prefix.dev/conda-forge
 mkdir -p build;
 cd build;
 export BUILD_PREFIX=$MAMBA_ROOT_PREFIX/xeus-cpp-wasm-build;


environment-wasm-build.yml
```diff
@@ -1,7 +1,7 @@
 name: xeus-cpp-wasm-build
 channels:
-  - https://prefix.dev/emscripten-forge-dev
+  - https://prefix.dev/emscripten-forge-4x
   - https://prefix.dev/conda-forge
 dependencies:
   - cmake
-  - emscripten_emscripten-wasm32==3.1.73
\ No newline at end of file
+  - emscripten_emscripten-wasm32==4.0.9
```
mcbarton, line 7: Why 4.0.9 and not 4.0.18, given that is the latest emsdk in Emscripten forges 4x branch? You mention in this comment https://github.com/compiler-research/xeus-cpp/issues/413#issuecomment-3544812490 that the 4x branch is now using 4.0.18, so wondering why this has been set to 4.0.9.

anutosh491: Cause emscripten-forge devs decided to ;)

Note : The version is still changeable. There's no change affecting us between 9 and 18

anutosh491: Cause emscripten-forge devs decided to ;)

Note : The version is still changeable. There's no change affecting us between 9 and 18

mcbarton: This comment is just a note so your aware. There may be no feature from emsdk 4.0.18 that xeus-cpp is using, but using emsdk 4.0.18 makes CppInterOp stall when running its browser tests in the ci. Haven't investigated locally yet what is happening to make it stall.

anutosh491: Well when we choose a version on emscripten-forge, we need to ensure the version we're using is able to compile the toughest library to compile there is ... Which is scipy (10/10 toughness).
Other libraries like xeus-cpp/cppinterop are easy to compile (2/10 toughness imo) (so if we can compile scipy through a version we for sure can compile our packages here)

That's probably the only decent explanation I have for now as to why we're going ahead with the version we're going ahead with.

environment-wasm-host.yml
```diff
@@ -1,6 +1,6 @@
 name: xeus-cpp-wasm-host
 channels:
-  - https://prefix.dev/emscripten-forge-dev
+  - https://prefix.dev/emscripten-forge-4x
```
mcbarton, line 3: @vgvassilev @JohanMabille @SylvainCorlay Do we really want this merged in before the 4x branch on Emscripten forge is considered ready to be merged in its main?

Asking since this was my thinking when I opened up my PR in CppInterOp updating it to using emsdk 4.0.9 (I put it on a checklist here https://github.com/compiler-research/CppInterOp/pull/753#issuecomment-3554559357 )

mcbarton: If we are not waiting for the 4x branch of Emscripten forge to be merged in main, I would like a chance to revert the CppInterOp PR back to 4.0.9, so its ready for the change too. Will try and do this change this evening, but may end up being tomorrow evening.

mcbarton: Another reason I am skeptical of this change going in before the 4x branch is ready is because I am told the emsdk version it is planning to settle on in this in flux, so we could up having to change emsdk again soon if its decided that a newer emsdk version 4 will end up being used when the 4x branch is ready.

anutosh491: All other kernels already migrated to 4-x and had their respective releases too.

We can change the channel link anytime so yeah no need to wait !

anutosh491: All other kernels already migrated to 4-x and had their respective releases too.

We can change the channel link anytime so yeah no need to wait !

mcbarton: @anutosh491 Please don't mark this as resolved. I would like the feedback from @vgvassilev @JohanMabille @SylvainCorlay first with regards to my issue of merging against an experimental branch of emscripten forge. If I felt your response answered my comment to my satisfaction I would have marked it as resolved, and is why I unresolved it the first time.

anutosh491:
cc @martinRenou

I just tested the emscripten 4-x support PR (https://github.com/jupyterlite/xeus/pull/311) using latest jupyterlite-xeus 4.3.0

Works just like we would want :)

The latest commit removes pin on jupyterlite-core (which is anyways fetched with jupyterlite-xeus I think)

environment-wasm-host.yml
anutosh491: > first with regards to my issue of merging against an experimental branch of emscripten forge

I don't understand this comment of yours. 

When I might have addressed the 4-x branch as "experimental", it means it's experimental with respect to building scipy and other rust packages but we already have all build relevant to us (llvm/cppinterop etc) on the branch already. Probably was experimental to us when the above packages were not built but now they are and its not experimental for us anymore.

Also, the branch might be experimental but the channel is a public channel and that's what we're putting to use. It already hosts a xeus-cpp build https://prefix.dev/channels/emscripten-forge-4x/packages/xeus-cpp which is built with the relevant changes from this PR put as patches on the recipe and works perfectly fine. Try fetching it and give it a shot.

There are technically about 300+ packages on main(3-x) and only after all move to 4-x we can say the "experiment" is over but how does that prevent us from using the channel ?

JohanMabille: > Do we really want this merged in before the 4x branch on Emscripten forge is considered ready to be merged in its main?

If all the packages required for the build and tests of the xeus-cpp wasm kernel are available, I don't see what would prevent us from doing so. And should there be any problem, reverting to emscripten-forge-dev is a one-line change, so this should not be a blocker.

mcbarton: @JohanMabille My suggested way forward to avoid this being a blocker is that is goes in, but if at any point xeus-cpps ci breaks due to a change in the experimental 4x branch, we immediately revert back to the dev branch

On a side I note I have tested this PR now against CppInterOps 4.0.9 update PR last night and it passes, so that is no longer a blocker assumming no more changes are made to the PR.

anutosh491: > but if at any point xeus-cpps ci breaks due to a change in the experimental 4x branch, we immediately revert back to the dev branch

We have a working combination right ?
That means we can pin the build number right ? 
Why do we need to revert anything in the first place ?

And any change would only affect us if we change the recipes for the dependencies involved in our case. So we should be good !

mcbarton: Basically your saying you can pin the build number so nothing should break in the future (in which case no reversion would happen). All I am saying if this turns out not to be true, and the experiment branch breaks our ci in the future (for example someone changes a recipe and it breaks) then we should revert back to the dev branch immediately. The one cavity is if you can fix it lets say by pinning a dependency to before the break then we don't revert.

anutosh491: I don't think I get what you say and to be honest, I'm out of patience to explain any better.

Why would anything break when I'm literally managing any change to our dependencies 🙃 ?

RESOLVING !

martinRenou:
> With jupyterlite-xeus 4.2.2, I see

Even though it "looks" like it works, it's quite broken as well. If you look carefully the logs you'll see that it switches all of your libraries to the emscripten-forge-dev channel (goes back to emscripten-abi 3.x).

But 4.2.2 is not supposed to work with emscripten 4.x so I won't fix it.

anutosh491:
> Is this a local jupyterlite build?

Nope, just following the readme (where we setup the xeus-lite-host env)

1) So the changes in this PR are also added as [patches](https://github.com/emscripten-forge/recipes/tree/emscripten-4x/recipes/recipes_emscripten/xeus-cpp/patches) on emscripten-forge's 4-x branch. So can fetch xcpp.wasm/js/data from here https://prefix.dev/channels/emscripten-forge-4x/packages/xeus-cp I assume

2) But otherwise check the environment-wasm-build.yml and environment-wasm-host.yml in this PR. No change apart from 
from changing the channel and the emscripten version

3) And yeah then 
```
micromamba create -n xeus-lite-host jupyter_server jupyterlite-xeus -c conda-forge
jupyter lite serve --XeusAddon.prefix=$PREFIX
```

anutosh491:
Also out of curiosity I tried mamba list and I don't see the channel (4-x) being mentioned 

image

Instead of something like the following on our main deployment 

image

martinRenou:
Mmh no that's not what you're saying sorry

JohanMabille (approved):

PR #422, opened 2025-11-30
Drop smallpt example notebook

mcbarton:
In this comment https://github.com/compiler-research/xeus-cpp/pull/415#issuecomment-3562772106 and on Discord @anutosh491 mentioned that the smallpt example isn't really adding much value to the repo, and is likely to be dropped before the next release. This PR removes the smallpt notebook from the repo, and any reference to it in the documentation.


anutosh491:
Drop all the SDL changes from Cmake too I guess!

mcbarton:
> Drop all the SDL changes from Cmake too I guess!

I did wonder whether to do this. I have done this now.

PR #426, opened 2025-12-04
Add tinyraytracer notebook

anutosh491:
# Description

As the smallpt notebook was removed, I'm taking this chance to add a better example imo here.

1) We dropped the SDL support along with smallpt which is great as SDL wasn't contributing much and we also reduce the size of xcpp.data so yeah faster loading. This doesn't mean we can't have the smallpt example in the notebook. Recreating it should be trivial.
2) But smallpt takes about 7 hours to generate a crystal clear image, and we were running it for about 2 mins to have a very poor image.
3) I've replaced it with an equally popular choice for ray tracing taught as a part of a university course (https://github.com/ssloy/tinyraytracer/wiki). Much better image in just 15 seconds and plain C++

image

## Type of change

- [ ] Bug fix
- [x] New feature


mcbarton:
My comments on this PR are

1) The notebook needs adding to the deployment and documentation

2) If you look at the faq for license the ray tracer example has here https://www.wtfpl.net/faq/ , you'll see it says 'The WTFPL lets you relicense the work under any other license.' . To me this says we can have the code in the repo under the bsd 3 of the rest of repo, so may not need to licence file added. This is not legal advice though. Just thought I'd bring it up so others can discuss.

mcbarton:
Just approved but I think it should be 'Tiny Ray Tracer using xeus-cpp' and not 'Tiny Ray Tracer in xeus-cpp'. Feel free to merge though if your ok with what it is.

anutosh491:
> Just approved but I think it should be 'Tiny Ray Tracer using xeus-cpp' and not 'Tiny Ray Tracer in xeus-cpp'. Feel free to merge though if your ok with what it is.

We had "smallpt in xeus-cpp" if you remember, so probably I just used that :~)

PR #436, opened 2026-01-26
Update xeus-cpp to CppInterOp 1.8.0

### LINKED ISSUE #186: Regarding pinning cxx-compiler to version 17
cxx-compiler 18 is out and like most xeus kernels we should be using it, but it seems like some tests fails on the ubuntu Ci when that is done (https://github.com/compiler-research/xeus-cpp/actions/runs/10836230049/job/30069522732)

https://github.com/compiler-research/xeus-cpp/blob/f98cca33cc448f3379e6350e21d3d8063a12b8d9/environment-dev.yml#L8

Look into why this fails and get rid of the pin.

mcbarton:
This PR gets xeus-cpp ready for the impending release of CppInterOp.

Fixes #186

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -5,13 +5,13 @@ dependencies:
   # Build dependencies
   - make
   - cmake
-  - cxx-compiler=1.7.0
+  - cxx-compiler
   # Host dependencies
   - xeus>=5.0.0
   - xeus-zmq<4.0
   - nlohmann_json
   - nlohmann_json-abi
-  - CppInterOp
+  - CppInterOp>1.7.0
   - pugixml
   - cpp-argparse
   # Test dependencies
--- a/environment-wasm-host.yml
+++ b/environment-wasm-host.yml
@@ -7,7 +7,7 @@ dependencies:
   - nlohmann_json-abi
   - xeus-lite
   - xeus
-  - CppInterOp
+  - CppInterOp>1.7.0
   - cpp-argparse
   - pugixml
   - doctest
--- a/include/xeus-cpp/xinterpreter.hpp
+++ b/include/xeus-cpp/xinterpreter.hpp
@@ -16,7 +16,7 @@
 #include <string>
 #include <vector>
 
-#include "clang/Interpreter/CppInterOp.h" // from CppInterOp package
+#include <CppInterOp/CppInterOp.h>
 
 #include <nlohmann/json.hpp>
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,9 +44,6 @@ jobs:
           - name: osx15-x86
             os: macos-15-intel
             micromamba_shell_init: bash
-          - name: osx14-arm
-            os: macos-14
-            micromamba_shell_init: bash
           - name: osx15-arm
             os: macos-15
             micromamba_shell_init: bash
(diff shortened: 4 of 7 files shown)


environment-dev.yml
```diff
@@ -5,13 +5,13 @@ dependencies:
   # Build dependencies
   - make
   - cmake
-  - cxx-compiler=1.7.0
+  - cxx-compiler
```
anutosh491, line 8: Wow super happy this works. How does it though ?

We made the pin cause we weren't able to address this issue correct ? https://github.com/compiler-research/CppInterOp/issues/422

And as far as I remember cxx-compiler 1.11.0 still wasn't able to address it (and there is no release out after that)

Maybe I'm missing something !

environment-wasm-host.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   - nlohmann_json-abi
   - xeus-lite
   - xeus
-  - CppInterOp
+  - CppInterOp>1.7.0
```
anutosh491, line 10: Why the pin ? 

is it because xeus-cpp would probably force wasm exceptions on master (I say probably cause Sylvain had some things to say about this, I'll be talking to him about it soon) and hence we also want the cppinterop version that is also built with wasm exceptions by default ?

vgvassilev: Yeah, and we have moved the CppInterOp.h file in a new place and we include it from there.

include/xeus-cpp/xinterpreter.hpp
```diff
@@ -16,7 +16,7 @@
 #include <string>
 #include <vector>
 
-#include "clang/Interpreter/CppInterOp.h" // from CppInterOp package
+#include <CppInterOp/CppInterOp.h>
```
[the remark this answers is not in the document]
vgvassilev: That seems legit, can you remove it from here?

include/xeus-cpp/xinterpreter.hpp
```diff
@@ -17,7 +17,6 @@
 #include <vector>
 
 #include <CppInterOp/CppInterOp.h>
-
```
vgvassilev, line 20: This did not remove the included header file.

mcbarton: Did it manually this time rather than relying on the bot.

environment-dev.yml
mcbarton: No idea why we can suddently remove the pin. Vassil suggested I try it after we had issues with getting the macos jobs to pass, and it just worked fine without the pin.

include/xeus-cpp/xinterpreter.hpp
mcbarton: Had to add it back after the tests failed when I removed the header. Didn't look at why.

anutosh491:
I was waiting for your fix on the feedstock "the llvm 21 fix" to land and then we could have had a build against that in the CI and then merged. But ig its fine

environment-dev.yml
```diff
@@ -5,13 +5,13 @@ dependencies:
...
   # Host dependencies
   - xeus>=5.0.0
   - xeus-zmq<4.0
   - nlohmann_json
   - nlohmann_json-abi
-  - CppInterOp
+  - CppInterOp>1.7.0
```
SylvainCorlay, line 14: In don't understand the strict upper bound on 1.7.0. 

Shouldn't it be a non-strict on 1.8?

PR #440, opened 2026-02-06
Make exception handling models better configurable for the wasm kernel

anutosh491:
# Description

This PR makes the wasm kernel better configurable towards exception handling model (native wasm exceptions or JS based exceptions or any others)

## Type of change

- [x] Bug fix
- [ ] New feature


CMakeLists.txt
```diff
@@ -83,6 +83,8 @@ if(EMSCRIPTEN)
     set(XEUS_CPP_USE_SHARED_XEUS_CPP OFF)
     # ENV (https://github.com/emscripten-core/emscripten/commit/6d9681ad04f60b41ef6345ab06c29bbc9eeb84e0)
     set(EMSCRIPTEN_FEATURES "${EMSCRIPTEN_FEATURES} -s \"EXPORTED_RUNTIME_METHODS=[ENV']\"")
+    set(XEUS_CPP_WASM_KERNEL_EXTRA_FLAGS "-fwasm-exceptions" CACHE STRING "Extra compiler/linker flags for wasm builds")
```
anutosh491, line 86: No need to have `sSUPPORT_LONGJMP=wasm` as it is enabled by default based on the EH model (wasm or emscripten)
 
 https://emscripten.org/docs/porting/setjmp-longjmp.html#c-setjmp-longjmp-support
 
 Can still mention it, if we want !

mcbarton (changes requested):
My understanding is that we want this change so xeus-cpp can remain compatible with the emsdk 3 branch on Emscripten forge which doesn't use wasm exceptions. So that we know the build is agnostic to the EMCC_CFLAGS from whatever emsdk version that user uses from emscripten forge I think this environment variable should be unset in the ci to stop it overwriting whatever the user has set for XEUS_CPP_WASM_KERNEL_EXTRA_FLAGS (probably want it unset in build instructions too). This should also allow CppInterOp to drop references to EMCC_CFLAGS when building xeus-cpp which it currently needs to do.

I would also like to be able to test this PR against CppInterOps ci.

One final remark, which I honestly don't know the answer to, does this test need to change depending on what exception model you use https://github.com/anutosh491/xeus-cpp/blob/3299d5c599fa220466c7751c4af279f4c6e22678/test/test_interpreter.cpp#L145 ?

CMakeLists.txt
mcbarton, line 86:
```suggestion
    set(XEUS_CPP_WASM_KERNEL_EXTRA_FLAGS "-fwasm-exceptions -sSUPPORT_LONGJMP=wasm" CACHE STRING "Extra compiler/linker flags for wasm builds")
```

If you look at the Windows Emscripten jobs in this CppInterOp PR https://github.com/compiler-research/CppInterOp/pull/803/checks which tests this xeus-cpp PR, you'll see that it has undefined errors for symbols such as `__wasm_longjmp `. Although `-sSUPPORT_LONGJMP=wasm` seems to be brought it automatically with `-fwasm-exceptions` on Unix systems, it is not being done automatically on Windows systems (probably some bug in Emscripten). With this change the CppInterOp ci should pass for the Emscripten build on all systems.

mcbarton (dismissed):

CMakeLists.txt
anutosh491: How does that make any sense ? Aren't we cross compiling here for the emscripten-wasm32 target ? 

If it works of any platform X, it should work on platform Y. 

> -sSUPPORT_LONGJMP=wasm seems to be brought it automatically with -fwasm-exceptions

Also I remember going through the source code for this upstream. Its a simple check, nothing to do with any platform is what I remember.

Check the windows CI for cppinterop if that's concerning. If you can't figure it out, I'll add this !

mcbarton: This has been set to resolved, without actually resolving. Please apply this change if you would like to resolve this comment. Without this change CppInterOp cannot drop the need for EMCC_CFLAGS when building xeus-cpp, since `-sSUPPORT_LONGJMP=wasm` is not being brought in automatically with `-fwasm-exceptions` on Windows.

SylvainCorlay (approved):

PR #443, opened 2026-02-11
Drop -march=native being a default flag when building xeus-cpp

mcbarton:
When building xeus-cpp `-march=native` shouldn't be a default flag, due to cases where xeus-cpp is cross compiled from one platform for a different platform. For example the osx 86 conda-forge package is built on a osx arm platform.


vgvassilev (approved):
LGTM!

PR #444, opened 2026-02-12
Reduce initial memory when building xeus-cpp tests to 64Mb

mcbarton:
xeus recently changed its initial memory flag to 32 Mb (here https://github.com/jupyter-xeus/xeus/pull/440/changes). This PR changes xeus-cpp to match.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -69,7 +69,7 @@ if(EMSCRIPTEN)
         PUBLIC "SHELL: -s ASSERTIONS=0"
         PUBLIC "SHELL: -s ALLOW_MEMORY_GROWTH=1"
         PUBLIC "SHELL: -s STACK_SIZE=32mb"
-        PUBLIC "SHELL: -s INITIAL_MEMORY=128mb"
+        PUBLIC "SHELL: -s INITIAL_MEMORY=64MB"
         PUBLIC "SHELL: --preload-file ${ESCAPED_SYSROOT_PATH}/include@/include"
         PUBLIC "SHELL: --preload-file ../${XEUS_CPP_DATA_DIR}@/share/xeus-cpp"
         PUBLIC "SHELL: --preload-file ../${XEUS_CPP_CONF_DIR}@/etc/xeus-cpp"


mcbarton:
> Interesting data point cc @martinRenou
> 
> Also what's the error ? Do you have some logs

The only logs I have to go by at the moment is the ci output (and can compare to main to see how output changes). We can probably get more if we run the browser tests locally, and inspect the console as they run.

test/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ if(EMSCRIPTEN)
         PUBLIC "SHELL: -s ASSERTIONS=0"
         PUBLIC "SHELL: -s ALLOW_MEMORY_GROWTH=1"
```
anutosh491, line 70: Basically what I conveyed through the edit section is I think starting with less initial memory and thinking that the allow memory growth flag would do the job might not work (I guess we can even drop this) .

martinRenou: I'm 99% sure `ALLOW_MEMORY_GROWTH` does the job. If it was not doing the job, xeus-python would simply not work (some Python environments we feed it are ~200MB).

I'd be interested to see an error message or something.

anutosh491: Okay yeah, let's see an error message first

mcbarton: image
Ruuning the tests from this PR locally in a Chrome browser, you get the following error message (see screenshot above showing the error message from the console)

```
Uncaught (in promise) RuntimeError: WebAssembly.instantiate(): data segment 0 is out of bounds (offset 34022176, length 18120, memory size 33554432)
```

martinRenou: Maybe it cannot grow memory during the initialization step, and if it needs more to initialize, it cannot?

Could you try again doubling that size?

Thanks for trying that out BTW

anutosh491: Ahhh as per my experience a "data segment 0 is out of bounds" error means 

that the allocated WebAssembly memory is too small for the data the module needs to initialize

anutosh491: Yeah doubling might just work. Cause offset + length is just beating memory size (34.... + 18... > 33....)

mcbarton: Doubling the initial memory to 64Mb allowed the browser tests to run when I tried locally in Chrome. Updated this PR to set the initial memory to 64Mb, so hopefully the ci should pass now.

martinRenou: So you were right calling that `ALLOW_MEMORY_GROWTH` doesn't do the job, but it doesn't do the job at initialization.

We should fix upstream xeus.

It seems to me pyodide has a very specific number: https://github.com/pyodide/pyodide/blob/1776000d7e64f587a9b714558b8a6e27238ddd91/Makefile.envs#L284

They probably figured out exactly how much it takes to initialize it.

anutosh491: > So you were right calling that ALLOW_MEMORY_GROWTH doesn't do the job, but it doesn't do the job at initialization.

Yeah we need to hit the correct amount at initialization.

> We should fix upstream xeus.
> It seems to me pyodide has a very specific number: https://github.com/pyodide/pyodide/blob/1776000d7e64f587a9b714558b8a6e27238ddd91/Makefile.envs#L284

How do they do this ? Trial and Error ? Confused how you hit a perfect number.

> We should fix upstream xeus.

Yeah this is like a good experiment here ! Do you think so too ? 

Can we extrapolate the finding from trying to build test_xeus_cpp.js to xcpp/xpython.js through xeus ?

anutosh491: cc @mcbarton do you want to try an exact figure here ? 😄 

`offset + length = 34,022,176 + 18,120 = 34,040,296`

Possibly try 35MB !? (or even `34,040,300` for that matter)

mcbarton: I tried locally. 40Mb was the smallest initial memory where the tests would run. 39Mb they failed to run. Not sure if it will be the same for xeus.

anutosh491: Hmm, okay. I guess let's just do 64 for being safe :\

 We might see crashes otherwise if let's say the test file doubles in size in the future ig.

mcbarton: Reverted initial memory in this PR back to 64Mb. 

The only way I can think you can see if this value works for xeus would be to add Emscripten tests to it. If you have browser tests using emrun like xeus-cpp, then the ci should fail if the value is not large enough.

martinRenou: 64Mb sounds good to me 👍🏽 we're already dividing by two the initial memory which is a win IMO.

The xeus change will go in a major release so we're kind of safe. I will make sure to test xeus-python with 64MB.

Thanks again folks 👍🏽

anutosh491: Sounds good to me. Approving !

Thanks Martin for the help 🙌

PR #453, opened 2026-03-06
Update xeus-cpp to be compatible with xeus 6.0

mcbarton:
With the recent release of xeus 6.0 the ci for xeus-cpp is failing on main due to the breaking changes made to xeus.

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -325,9 +325,9 @@ __get_cxx_version ()
         return result;
     }
 
-    nl::json interpreter::shutdown_request_impl(bool restart)
+    nl::json interpreter::shutdown_request_impl(bool /*restart*/)
     {
-        return xeus::create_shutdown_reply(!restart);
+        return xeus::create_shutdown_reply(false);
     }
 
     nl::json interpreter::interrupt_request_impl()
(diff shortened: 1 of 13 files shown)


anutosh491:
I would have just waited for @JohanMabille to make this change cause he's the most educated about this the update to xeus 6 😬

But yeah let's ask him to review this right now as you've already made it !

src/xinterpreter.cpp
```diff
@@ -325,9 +325,14 @@ __get_cxx_version ()
         return result;
     }
 
-    void interpreter::shutdown_request_impl()
+    nl::json interpreter::shutdown_request_impl(bool restart)
     {
-        restore_output();
+        return xeus::create_shutdown_reply(!restart);
```
JohanMabille: You should pass false here, unless xcpp implements some restart mechanism.

JohanMabille (changes requested):
The idea is also to always use [helper functions](https://github.com/jupyter-xeus/xeus/blob/main/include/xeus/xhelper.hpp) to create the response messages instead of building them adhoc.

Also the kernelspec should be updated to contain the new field "kernel_protocol_version" and with value "5.6.0".

src/xinterpreter.cpp
mcbarton: That is what I originally had, copying xeus-python but then the ci fails due to having the unused restart variable. This was my workaround. You pass through true to restart, and then it passes through the negation (false). Do you have an alternative suggested way of doing this where the restart variable is used?

JohanMabille: You can simply comment the variable name (or remove it) so that you don't have the warning about the unused variable:
```cpp
nl::json interpreter::shutdown_request_impl(bool /*restart*/)
```

PR #455, opened 2026-03-11
Fixed inspect_request implementation

JohanMabille:
Fixes the implementation of inspection.

### DIFF
--- a/src/xinspect.cpp
+++ b/src/xinspect.cpp
@@ -8,9 +8,11 @@
 
 #include <filesystem>
 #include <fstream>
+#include <iostream>
 #include <regex>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "xeus/xhelper.hpp"
 
--- a/src/xinspect.hpp
+++ b/src/xinspect.hpp
@@ -14,6 +14,8 @@
 
 #include <pugixml.hpp>
 
+#include <nlohmann/json_fwd.hpp>
+
 #include "xeus-cpp/xpreamble.hpp"
 #include "xeus-cpp/xutils.hpp"
 
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -7,6 +7,8 @@
  * The full license is in the file LICENSE, distributed with this software.         *
  ************************************************************************************/
 
+#include <regex>
+
 #include "xeus/xhelper.hpp"
 #include "xeus/xsystem.hpp"
 
@@ -264,8 +266,6 @@ __get_cxx_version ()
 
     nl::json interpreter::inspect_request_impl(const std::string& code, int cursor_pos, int /*detail_level*/)
     {
-        nl::json kernel_res;
-        std::string exp = R"(\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)";
         std::regex re(R"((\w*(?:\:{2}|\<.*\>|\(.*\)|\[.*\])?)(\.?)*$)");
 
         std::smatch inspect_request;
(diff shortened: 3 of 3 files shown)


src/xinspect.cpp
```diff
@@ -257,7 +228,19 @@ namespace xcpp
         std::regex re(spattern + R"((.*))");
         std::smatch to_inspect;
         std::regex_search(code, to_inspect, re);
-        inspect(to_inspect[1], kernel_res);
+        std::string result = inspect(to_inspect[1]);
+        if (result.empty())
+        {
+            kernel_res = xeus::create_error_reply("No documentation found");
+        }
+        else
```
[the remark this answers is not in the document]
JohanMabille: The previous message did no include the queried expression (it is an error message, the parent_header is enough to know wat failed).

test/test_interpreter.cpp
```diff
@@ -286,13 +284,7 @@ TEST_SUITE("execute_request")
 
 TEST_SUITE("inspect_request")
 {
-#if defined(__EMSCRIPTEN__)
-    TEST_CASE("good_status"
-            * doctest::should_fail(true)
-            * doctest::description("TODO: Currently fails for the Emscripten build"))
-#else
-    TEST_CASE("good_status")
-#endif
```
anutosh491, line 295: Nice !!

We obviously has an inspect request working with xeus-cpp-lite (can be seen in the demo notebook) but somehow the implementation was going wrong for the wasm use case and hence this test was failing.

```
===============================================================================
[doctest] test cases: 47 | 47 passed | 0 failed | 0 skipped
[doctest] assertions: 73 | 73 passed | 0 failed |
[doctest] Status: SUCCESS!
```

Now this passes in the CI which is already a great sign for this PR.

src/xinspect.cpp
```diff
@@ -257,7 +228,19 @@ namespace xcpp
...
+        {
+            auto payload = nl::json::array();
+            payload[0] = nl::json::object(
+                {{"data", build_inspect_data(result)}, {"source", "page"}, {"start", 0}}
+            );
+            kernel_res = xeus::create_successful_reply(payload);
+        }
```
anutosh491, line 245: Guessing we would also like to change `apply` from xpreamble to only accept a string for `code` and not need a kernel_res passed to callback eventually ?

https://github.com/compiler-research/xeus-cpp/blob/4d4d2b2c6d61c67a82486b8940bd9a5731baa1e1/include/xeus-cpp/xpreamble.hpp#L32

I see this being used in the execute request (and I guess we can use a helper here too ?)
https://github.com/compiler-research/xeus-cpp/blob/4d4d2b2c6d61c67a82486b8940bd9a5731baa1e1/src/xinterpreter.cpp#L141-L142

anutosh491: Offcourse this would do too, I just thought of the above refactor for sake of completetion (as to how inspect is called in the inspect request)

JohanMabille: > Guessing we would also like to change apply from xpreamble to only accept a string for code and not need a kernel_res passed to callback eventually ?

We could, but that would be a breaking change and I don't know how many implementations inherit from this class, so I prefered to be conservative here.

JohanMabille: > I see this being used in the execute request (and I guess we can use a helper here too ?)

The implementation of apply already uses a helper to build the kernel_res. I'll plug other helpers in interpreter in a dedicated PR.

anutosh491: Yeah makes sense.

src/xinspect.cpp
```diff
@@ -60,71 +64,59 @@ namespace xcpp
...
+            return "";
         }
-        return result;
-    }
 
-    std::pair<bool, std::smatch> is_inspect_request(const std::string& code, const std::regex& re)
-    {
-        std::smatch inspect;
-        if (std::regex_search(code, inspect, re))
+        nl::json read_tagconfs(const char* path)
```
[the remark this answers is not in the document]
JohanMabille: header already included in xinspect.hpp ...

PR #475, opened 2026-04-12
Set XEUS_CPP_OMP to -fopenmp (don't require openmp to be available to build time to set)

mcbarton:


vgvassilev (approved):
LGTM! Do we have OMP tests in tree?

mcbarton:
> LGTM! Do we have OMP tests in tree?

We currently have a single test here https://github.com/compiler-research/xeus-cpp/blob/978c52c51e2d46e0002966ae106e0443bb21bbc2/test/test_xcpp_kernel.py#L246 . It creates 2 threads, and gets them to output their thread number in descending order. It was added just to test basic openmp functionality. I will work on adding more tests in a future PR.

PR #479, opened 2026-04-22
Account for change of name of CppInterOp target (clangCppInterOp to CppInterOp)

mcbarton:
This PR is to stop https://github.com/compiler-research/CppInterOp/pull/882 being a breaking change as far as xeus-cpp is concerned. The library is changing name. This PR intends to extract the target name from the library, and configure xeus-cpp accordingly. 

There is probably be a better way to do this, but this is what I came up with.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,13 @@ if (NOT TARGET xeus AND NOT TARGET xeus-static)
 endif ()
 
 find_package(CppInterOp ${CppInterOp_REQUIRED_VERSION} REQUIRED)
+
+# Extract filename without extension
+get_filename_component(_name "${CPPINTEROP_LIBRARIES}" NAME_WE)
+
+# Remove "lib" prefix
+string(REGEX REPLACE "^lib" "" CPPINTEROP_TARGET "${_name}")
+
 if(CppInterOp_FOUND)
   message(STATUS "Found CppInterOp: config=${CPPINTEROP_CMAKE_DIR} dir=${CPPINTEROP_INSTALL_PREFIX} (found version=${CPPINTEROP_VERSION} compatible with Clang ${CPPINTEROP_LLVM_VERSION_MAJOR}.x)")
 endif()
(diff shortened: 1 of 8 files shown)


anutosh491 (commented):
Wait i'm confused here.
We have this right now 

https://github.com/compiler-research/xeus-cpp/blob/88a82c50188686dec7e36ee82e1597700c920ad5/CMakeLists.txt#L93

And cppinterop 2.0 would somehow be a breaking one I'm guessing and we would want to update the above in cmake to be 2.0 and hence that eliminates the chance of us using  `clangCppInterOp` right ? (as we compulsarily require 2.0)

Once we move to 2, we would anyways have to change to `libCppInterOp.so`... so I'm not sure why we simply don't just that ?

mcbarton:
> Wait i'm confused here. We have this right now
> 
> https://github.com/compiler-research/xeus-cpp/blob/88a82c50188686dec7e36ee82e1597700c920ad5/CMakeLists.txt#L93
> 
> And cppinterop 2.0 would somehow be a breaking one I'm guessing and we would want to update the above in cmake to be 2.0 and hence that eliminates the chance of us using `clangCppInterOp` right ? (as we compulsarily require 2.0)
> 
> Once we move to 2, we would anyways have to change to `libCppInterOp.so`... so I'm not sure why we simply don't just that ?

This is one way to do it, and one way I suggested tackling this problem in the CppInterOp PR (e.g. let CppInterOps depoyment be broken for a commit, and then do a change to the library name in xeus-cpp after release, breaking compatibility with previous cppinterop versions). The method in this PR makes it so the name change is not a breaking change. CppInterOp 2.0 is not necessarily breaking despite the major version change. Vassil chose to be 2.0 as the root cppyy migration may require breaking changes (so far no changes other than this name change would be breaking for xeus-cpp)

anutosh491:
I think if we would make the cmake change, we won't need this I think !

And I guess we should just do it like we do the update everytime there is a cppinterop release.

PR #481, opened 2026-04-27
fix: include <sstream> in audio cell of demo notebook

### LINKED ISSUE #480: Running cell 15 from xeus-cpp.ipynb demo notebook fails
Image

We just need to include `sstream` here 

https://github.com/compiler-research/xeus-cpp/blob/88a82c50188686dec7e36ee82e1597700c920ad5/test/Notebooks/xeus-cpp.ipynb#L360

mvanhorn:
Closes #480.

Adds `#include <sstream>` to the audio cell of `test/Notebooks/xeus-cpp.ipynb`. The cell uses `std::stringstream` (`au::audio::m_buffer`) without the corresponding header, so toolchains that don't transitively include `<sstream>` via `<string>`/`<fstream>` fail to compile the cell, which is what produced the failure in #480.

The companion wasm demo notebook (referenced by @anutosh491 in the issue) already has the include — this brings the desktop demo in line.

Single-line, JSON-valid edit (the notebook still parses cleanly with `json.load`).

### DIFF
--- a/test/Notebooks/xeus-cpp.ipynb
+++ b/test/Notebooks/xeus-cpp.ipynb
@@ -358,6 +358,7 @@
    "source": [
     "#include <string>\n",
     "#include <fstream>\n",
+    "#include <sstream>\n",
     "\n",
     "#include \"nlohmann/json.hpp\"\n",
     "\n",


JohanMabille (approved):
Thanks for the fix!

PR #485, opened 2026-06-13
fix: replace iframe with link in build_inspect_data

### LINKED ISSUE #484: [Bug] inspect request failure
I've been seeing this since almost more than a month now 

Image

I see the same on the native kernel too

Image

mvanhorn:
## Summary

- Replaces the `<iframe>` in `build_inspect_data` (src/xinspect.cpp) with a plain `<a>` anchor that opens the documentation URL in a new tab
- Removes the now-unused `#pager-container` / `.xcpp-iframe-pager` CSS block
- Adds a unit test in `test/test_interpreter.cpp` that asserts the returned `text/html` contains an `<a href>` and no `<iframe>`

## Why this matters

Closes #484. cppreference.com migrated to MediaWiki, which sets `X-Frame-Options: DENY` by default as a deliberate security policy. This blocks the `<iframe>` pager in every browser (Chrome, Safari, JupyterLite), leaving the Shift+Tab inspect panel blank for all users. The maintainer confirmed that changing the cppreference server config is not feasible, so the fix must be on the xeus-cpp side. A clickable link is the simplest correct replacement - it requires no special permissions and works in all Jupyter environments.

Fixes #484

### DIFF
--- a/src/xinspect.cpp
+++ b/src/xinspect.cpp
@@ -103,6 +103,109 @@ namespace xcpp
             }
             return result;
         }
+
+        std::string escape_html_attribute(const std::string& value)
+        {
+            std::string result;
+            result.reserve(value.size());
+            for (char c : value)
+            {
+                switch (c)
+                {
+                    case '&':
+                        result += "&amp;";
+                        break;
+                    case '"':
+                        result += "&quot;";
+                        break;
+                    case '\'':
+                        result += "&#39;";
+                        break;
+                    case '<':
+                        result += "&lt;";
+                        break;
+                    case '>':
+                        result += "&gt;";
+                        break;
+                    default:
+                        result += c;
+                }
+            }
+            return result;
+        }
+
+        std::string build_cppreference_srcdoc(const std::string& inspect_result)
+        {
+            return R"(<!doctype html>
+<html>
+<head>
+<meta charset="utf-8">
+<base href=")" + escape_html_attribute(inspect_result)
+                   + R"(" target="_blank">
+<style>
+body { margin: 0; font-family: system-ui, sans-serif; }
+.xcpp-status { padding: 1rem; }
+main { overflow: auto; }
+img { max-width: 100%; }
+pre { overflow: auto; }
+table { border-collapse: collapse; }
+th, td { border: 1px solid #aaa; padding: .25rem .5rem; }
+</style>
+</head>
+<body data-documentation-url=")"
+                   + escape_html_attribute(inspect_result) + R"(">
+<p class="xcpp-status">Loading documentation… <a href=")"
+                   + escape_html_attribute(inspect_result) + R"(">Open cppreference</a></p>
+<main></main>
+<script>
+const documentationUrl = document.body.dataset.documentationUrl;
+const status = document.querySelector('.xcpp-status');
+const content = document.querySelector('main');
+const pageUrl = new URL(documentationUrl);
+const page = pageUrl.pathname.replace(/^\/w\//, '').replace(/^\//, '');
+const apiUrl = new URL('/mwiki/api.php', pageUrl);
+apiUrl.search = new URLSearchParams({
+    action: 'parse',
+    page,
+    prop: 'text',
+    format: 'json',
+    formatversion: '2',
+    origin: '*'
+});
+
+fetch(apiUrl)
+    .then((response) => {
+        if (!response.ok) {
+            throw new Error(`HTTP ${response.status}`);
+        }
+        return response.json();
+    })
+    .then((data) => {
+        if (!data.parse || !data.parse.text) {
+            throw new Error('Documentation was not returned');
+        }
+        content.innerHTML = data.parse.text;
+        content.querySelectorAll('[href], [src]').forEach((element) => {
+            ['href', 'src'].forEach((attribute) => {
+                const value = element.getAttribute(attribute);
+                if (value && !value.startsWith('#')) {
+                    element.setAttribute(attribute, new URL(value, documentationUrl).href);
+                }
+            });
+        });
+        content.querySelectorAll('a').forEach((link) => {
+            link.target = '_blank';
+            link.rel = 'noopener noreferrer';
+        });
+        status.remove();
+    })
+    .catch(() => {
+        status.firstChild.textContent = 'Unable to load documentation inline. ';
+    });
+</script>
+</body>
+</html>)";
+        }
     }
 
     std::string inspect(const std::string& code)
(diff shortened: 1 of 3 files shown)


anutosh491 (commented):
Curious as to how the deployment looks like (maybe share a screencast or an image ?)

We also have a solution here : https://github.com/compiler-research/xeus-cpp/issues/484#issuecomment-4743584395

mvanhorn:
xeus-cpp inspect pager: before / after

I could not capture this inside a live xeus-cpp deployment, both linked JupyterLite demos currently load Lab but expose no C++ kernel, so I reproduced the pager payloads directly in Chrome against en.cppreference.com. Same HTML, same origin behaviour.

The result changed my mind about this PR, so let me lead with that.

The blocker is `?action=purge`, not cppreference:

    $ curl -sI "https://en.cppreference.com/w/cpp/container/vector?action=purge" | grep -i x-frame
    x-frame-options: DENY

    $ curl -sI "https://en.cppreference.com/cpp/container/vector" | grep -i x-frame
    (nothing)

Article views send no `X-Frame-Options` at all today. Only the `?action=purge` URL does, and that suffix is appended by `build_inspect_data` itself. Pane 1 above is the current code, showing the blocked-document icon; pane 2 is the identical iframe with `?action=purge` removed, rendering the full `std::vector` page.

So there is a third option nobody in #484 has raised, and I think it beats both on the table: delete `?action=purge` and keep the iframe. One line, keeps documentation inline where @mcbarton wants it, and needs no mirror, no hosting and no CC-BY-SA determination.

Two caveats. `?action=purge` was presumably there to defeat MediaWiki caching, so dropping it means a stale page could be served; for read-only documentation that seems a fair trade, but it is your call. And these are the headers cppreference serves today, so if they add `X-Frame-Options` to article views later, the iframe breaks again and the link approach in this PR becomes right after all.

Happy to reshape #485 into the one-line removal if you agree, keeping the test but asserting the emitted src no longer carries the purge suffix.

Separately, unrelated to this PR: both https://compiler-research.github.io/xeus-cpp/lab/index.html and https://compiler-research.org/xeus-cpp-wasm/lab/index.html load Lab but offer no C++ kernel in the launcher or kernel picker, so notebooks open with "No Kernel". Might be worth a look.

mcbarton:
> Separately, unrelated to this PR: both https://compiler-research.github.io/xeus-cpp/lab/index.html and https://compiler-research.org/xeus-cpp-wasm/lab/index.html load Lab but offer no C++ kernel in the launcher or kernel picker, so notebooks open with "No Kernel". Might be worth a look.

@mvanhorn I tried both deployments and the kernels are there for me. 

In the past when the kernels have not appeared for me it was because I changed the config on my browser and that somehow effected it. No idea why. I just reset my browser config to the defaults and they appeared again. I say this not knowing if it will help you.

mcbarton:
I am for a technique which avoids opening the documentation in a new tab or window. I have been learning about accessibility in web pages recently, and `Opening new windows and tabs from a link only when necessary` is something which is advised (see https://www.w3.org/WAI/WCAG22/Techniques/general/G200)

anutosh491:
Hey @mvanhorn, 

We would love if the behaviour is as close as what we had originally i.e the docs render below the cell that calls for it !

If you're approach can get us to that, let's go for it.

mvanhorn:
Good news, I think this can get us back to docs rendering below the cell. Pushed 1253887.

The reason the original iframe broke is that cppreference sends `X-Frame-Options: DENY`, so no browser will frame the page directly. The way around it is to not frame the page at all, and instead fetch its content through cppreference's own MediaWiki API and render that inline:

- `build_inspect_data` now emits an iframe with `srcdoc` for `https://en.cppreference.com/w/` URLs. The document inside calls `/mwiki/api.php?action=parse&prop=text&origin=*` for the requested page and injects the returned HTML.
- The iframe is sandboxed with `allow-scripts allow-popups allow-popups-to-escape-sandbox` but deliberately **not** `allow-same-origin`. The documentation HTML therefore runs at an opaque origin and cannot touch the notebook page. That is also why the API call passes `origin=*`, which is MediaWiki's anonymous CORS mode and the reason this works from a null origin at all.
- Relative links and images are rewritten against the documentation URL so images and cross-references resolve, anchors get `target="_blank" rel="noopener noreferrer"`, and every interpolated URL is HTML-escaped.
- If the fetch fails for any reason, the frame degrades to the plain "Open cppreference" link rather than showing an empty box, so we are never worse off than the link-only version.
- Non-cppreference URLs keep the direct `src` path, since they have no `X-Frame-Options` problem.

Tests updated accordingly, including a case covering the non-cppreference branch.

One caveat on verification: I do not have the LLVM/Clang-REPL toolchain locally, so I have not done a full build. I did confirm the raw string literals lex cleanly under clang and that the embedded script passes `node --check`, but CI is the real check here. Also worth your judgement: this does make inspect depend on cppreference's API being reachable at render time. The fallback covers it, but if you would rather not take that dependency, say so and I will revert to the link.

anutosh491:
> Good news, I think this can get us back to docs rendering below the cell.

Lovely. That's exactly what I'd like to go back to!

anutosh491:
The tests fail. Do you know why ?

mcbarton:
> The tests fail. Do you know why ?

The tests are not failing. There is a Valgrind issue, but this exists on main too. The tests fail to build on Windows now due to an unresolved symbol error. It appears to be coming from the `nlohmann::json` abi package, but I cannot tell beyond that. It looks like the sort of think which requires an upstream patch, but I don't know much about Windows.

mvanhorn:
> The tests fail. Do you know why ?

Yes -- and it's ours, not upstream. Sorry @mcbarton, I think the `nlohmann::json` trail is a red herring.

The Windows jobs pass on `main` and fail only here, which rules out a pre-existing ABI problem. The cause is in this repo's own header: `src/xinspect.hpp` declares

```cpp
std::string inspect(const std::string& code);
nl::json build_inspect_data(const std::string& inspect_result);
```

with no `XEUS_CPP_API`, while every type around them has it (`node_predicate`, `class_member_predicate`, `xintrospection`). That macro is `__declspec(dllexport)` / `__declspec(dllimport)` on Windows and expands to nothing everywhere else, so those two symbols were never in the DLL's export table. Nothing outside the library called them, so it never surfaced -- until this PR added a test that calls `xcpp::build_inspect_data` from the test binary, which is exactly an unresolved external on MSVC and invisible on Linux and macOS.

Fixed in `daf4ea0` by marking both declarations `XEUS_CPP_API`. I can't reproduce an MSVC link locally, so the Windows jobs here are the real check.

The other two failures do look pre-existing: `ubu24-analyzers` and `ubu24-arm-analyzers` fail on `main` at `4c7ab74` as well, so I don't think they're this PR's to fix. @mcbarton's Valgrind read matches that.

mcbarton:
image
I just ran the notebook in the Jupyter Lite preview for this PR, and it does not have the previous behaviour. A message saying you cannot open the documentation inline and providing a link to the cppreference page which opens in a new tab is not what we want.

mcbarton (changes requested):
Putting a `request changes` on this PR until the same behaviour we had previously is observed with respect to the documentation rendering.

mvanhorn:
Thanks for trying the preview, @mcbarton, that was the right test and it caught a real bug. The srcdoc script was fetching `/mwiki/api.php` on cppreference, which now answers 404 "No such action", so the fetch failed every time and the pager fell back to the plain link you saw. The working endpoint is `/api.php`; it returns the parsed page with `Access-Control-Allow-Origin: *`, so the sandboxed iframe can load it. Fixed in c4a056b (path plus the matching test assertion). @anutosh491, with that the docs render below the cell again rather than in a new tab, which is the behaviour you asked for; I could not run the JupyterLite build locally, so a second look at the preview once CI publishes it would be welcome.

PR #487, opened 2026-08-26
Replace prepare-dell part of ci with ci-workflows action

mcbarton:


mcbarton:
The changes in this PR make xeus-cpp line up with how CppInterOp and Clad do the prepare-dell part of the ci.

The failing jobs having nothing to do with the changes in this PR. The failures have been happening with the ci on main since around August 12th. Something happened around that time to introduce a Valgrind error in xeus-cpps tests.

